PR #5, opened 2020-02-25
TestFWCore Subdirectory with ExampleEventDataProducer

vvolkl:
Adds a test and example how to use the I/O components in the Framework.


vvolkl:
This moved to key4hep/k4-project-template.

PR #11, opened 2020-06-16
Link edm4hep to FWCore

vvolkl:
... so subdirectories linking to FWCore do not have to specify it explicitly.

### DIFF
--- a/FWCore/CMakeLists.txt
+++ b/FWCore/CMakeLists.txt
@@ -12,7 +12,7 @@ gaudi_install_python_modules()
 gaudi_add_library(FWCore
 		  src/*.cpp
                   INCLUDE_DIRS podio ROOT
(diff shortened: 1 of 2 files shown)


FWCore/CMakeLists.txt
```diff
@@ -13,7 +12,7 @@ gaudi_install_python_modules()
 gaudi_add_library(FWCore
 		  src/*.cpp
                   INCLUDE_DIRS podio ROOT
-                  LINK_LIBRARIES GaudiAlgLib GaudiKernel podio::podioRootIO ROOT
+                  LINK_LIBRARIES GaudiAlgLib GaudiKernel ROOT podio::podioRootIO EDM4HEP::edm4hepDict
```
andresailer, line 15: Does it need the "no-as-needed" flag?

PR #12, opened 2020-06-25
First Particle Gun

vvolkl:
Adding a Particle Gun, mostly in order to have something to test the Delphes Integration.


vvolkl:
At the moment failing because of #13.

vvolkl:
Not sure where to move this yet, but also will not stay here.

PR #18, opened 2020-11-04
Restructuring: FWCore->k4FWCore

vvolkl:
Also merges library the library and module. 

This fixes a collision with FWCore in FCCSW (to be removed).


vvolkl:
Hi @mirguest 

this will require some changes in depending code, which is mostly CEPCSW at the moment. Could you check that this is fine for you? It will make everything more consistent in the end.

The changes in CEPCSW needed, are, I believe:


* `s/#include "FWCore/DataHandle.h"/#include "k4FWCore/DataHandle.h"/g`
* `LINK_LIBRARIES ... FWCore ...` -> `LINK_LIBRARIES GaudiKernel k4FWCore`
* `INCLUDE_DIRS FWCore` -> `INCLUDE_DIRS k4FWCore`

To use Gaudi v35r0, unfortunately there will again be some code changes necessary, but this is not connected to this PR.

mirguest:
Hi @vvolkl ,
Sure, I will checkout your branch and test it with CEPCSW. 
Tao

vvolkl:
Splitting out the fixes to the ci into #20

mirguest:
Hi @vvolkl ,
Could you rename the python module?
    https://github.com/vvolkl/K4FWCore/tree/rename/k4FWCore/python/FWCore
Tao

vvolkl:
Hi Tao, good catch! I'll remove it alltogether, since the command line arguments are now fully handled by `k4run`

vvolkl:
I take this opportunity as well to migrate the overlay components to their own package.

mirguest:
Thanks @vvolkl .
My another question is in https://github.com/vvolkl/K4FWCore/blob/rename/CMakeLists.txt#L21, we still use `K4FWCore` as the project name. So do we also need to rename it?
Tao

k4FWCore/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+# Package: k4FWCore
+################################################################################
+gaudi_subdir(k4FWCore v1r0)
+
+find_package(podio)
+
+gaudi_install_scripts()
+
+
+gaudi_add_module(k4FWCore
```
mirguest, line 11: Hi @vvolkl , I have a trouble with this. In the previous version, there is `gaudi_add_library` and `gaudi_add_module` respectively. The first marco declares the public header, then the k4FWCore header directory will be installed. But in this branch, no header will be installed.
Could you have a look here? Thank you!
Tao

vvolkl: Oh, I forgot about that, but it looks like I can also install the headers for the module. Thanks!

vvolkl:
> Thanks @vvolkl .
> My another question is in https://github.com/vvolkl/K4FWCore/blob/rename/CMakeLists.txt#L21, we still use `K4FWCore` as the project name. So do we also need to rename it?
> Tao

Good point. For more consistency, I rename every instance of K4 to k4. I update my comment above.

mirguest:
@vvolkl Thank you very much!
Now I can build k4FWCore with Gaudi 33.1 without any problem.
The next issue is building CEPCSW with k4FWCore. I found following problem during cmake:
```
CMake Error at /cvmfs/cepcsw.ihep.ac.cn/prototype/releases/externals/97.0.2/Gaudi/cmake/GaudiProjectConfig.cmake:1626 (message):
  k4FWCore is a MODULE_LIBRARY: you cannot link against it
Call Stack (most recent call first):
  /cvmfs/cepcsw.ihep.ac.cn/prototype/releases/externals/97.0.2/Gaudi/cmake/GaudiProjectConfig.cmake:2097 (gaudi_resolve_link_libraries)
  /cvmfs/cepcsw.ihep.ac.cn/prototype/releases/externals/97.0.2/Gaudi/cmake/GaudiProjectConfig.cmake:2260 (gaudi_common_add_build)
  Digitisers/DCHDigi/CMakeLists.txt:20 (gaudi_add_module)


-- Configuring incomplete, errors occurred!
See also "/tmp/lint/key4hep/CEPCSW/build.97.0.2.x86_64-slc6-gcc8-opt/CMakeFiles/CMakeOutput.log".

```

vvolkl:
Hmmm, Indeed it seems that Gaudi doesn't allow linking module libraries. I'll revert to the previous split between k4FWCore and k4FWCorePlugins.

PR #35, opened 2021-03-02
Add an option to ignore the first N events in the input stream

zoujh:
Sometimes we want to debug an appointed event. And we need to read events from the middle of a file, rather than from the beginning. With this update, users can set the first event to read from the input stream.

### DIFF
--- a/k4FWCore/components/k4DataSvc.cpp
+++ b/k4FWCore/components/k4DataSvc.cpp
@@ -9,6 +9,7 @@ DECLARE_COMPONENT(k4DataSvc)
 k4DataSvc::k4DataSvc(const std::string& name, ISvcLocator* svc) : PodioDataSvc(name, svc) {
   declareProperty("inputs", m_filenames = {}, "Names of the files to read");
   declareProperty("input", m_filename = "", "Name of the file to read");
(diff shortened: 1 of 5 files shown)


vvolkl (changes requested):
Hi @zoujh,

thanks a lot, that is definitely a very helpful feature. Do you think you could add a test (I can help with the technicalities if you are not familiar with CTest)?

Also, I think the same thing could be achieved simply by changing the `EvtSel` option on the ApplicationMgr, but I will need to ask the Gaudi developers how exactly this works.

zoujh:
Hi @vvolkl,

I add a test script. It's my first time to deal with CTest, so please point it out if there is any incorrectness. Thank you!

And, of course, I am glad if there is a simple way to achieve this by the `EvtSel`.

vvolkl:
Hi @zoujh ,

thanks for the test, that looks very good! I opened an issue on gaudi, and indeed it seems like the functionality should be implemented in a custom EventSelector. I will take a look.

vvolkl:
Investigating the Gaudi option will likely take some more time, so I'll merge this now. Sorry for the delay!

PR #38, opened 2021-03-15
k4Run: change command line option for props to include the name

andresailer:
of the configurable and not just the name of the property

if there is more than one instance of a configurable (e.g., k4MarlinWrapper), one cannot otherwise set the option for them or if different configurables have the same property names.

BEGINRELEASENOTES
- k4run: change properties to include the name of the configurable: E.g. --Files --> --LcioEvent.Files

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/scripts/k4run
+++ b/k4FWCore/scripts/k4run
@@ -103,11 +103,12 @@ if __name__ == "__main__":
                           proptype = type(propvalue[0])
                         propnargs = "+"
                         
-                      parser.add_argument("--" + prop, type=proptype, help=props[prop][1],
+                      propName = conf.name() + '.' + prop
+                      parser.add_argument("--%s" % propName, type=proptype, help=props[prop][1],
                         nargs=propnargs,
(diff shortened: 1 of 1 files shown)


andresailer:
This should also solve the problem of duplicates
https://github.com/key4hep/k4FWCore/blob/da9e39c190bd6152ed908ebc56802febb829319d/k4FWCore/scripts/k4run#L84
I think

vvolkl:
Thanks Andre. Looks, good but I think I might add some logic to also add the "shortcut" of propname only  if there is only one configurable that provides a certain propname.

This is to keep backwards compatibility for one, and  because outside of the k4marlinwrapper use case,  configurables mostly appear in one instance.
A lot of their property names are also already explanatory enough to use on the command line, so this  adds a lot of keystrokes in that case, and when things are likely to be typed it's always good to keep it concise.

andresailer:
Thanks Valentin, sounds reasonable to me.

PR #39, opened 2021-03-23
[k4run] restore argument shortcuts where possible

vvolkl:
See discussion in #38 

The "shortcuts" are more tricky to work with than the full property names - if there are two shortcuts with the same name, the order of the configurables in the options file determines which one gets set by the shortcut. So where brevity is not important, it's probably better to use `--PodioOutput.filename` instead of `--filename`, for example. Still, as sometimes these will get typed, I think it's good to have.

BEGINRELEASENOTES
- add property arguments shortcuts where possible.

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/scripts/k4run
+++ b/k4FWCore/scripts/k4run
@@ -82,43 +82,34 @@ if __name__ == "__main__":
                   if not hasattr(props[prop][0], '__slots__'):
                     # argparse ArgumentError: used to catch when trying to add
                     # same argument twice
-                    try:                       
-                      propvalue = props[prop][0]
+                    propvalue = props[prop][0]
 
-                      # if it is set to "no value" it hasn't been touched in the options file
+                    # if it is set to "no value" it hasn't been touched in the options file
 
-                      if propvalue == conf.propertyNoValue:
-                         propvalue = conf.getDefaultProperty(prop) # thus get the default value
-                      proptype = type(props[prop][0])
-                      # if the property is a list of something, we need to set argparse nargs to '+'
-                      propnargs = "?"
-                      if proptype == list:
-                        # tricky edgecase: if the default is an empty list there is no way to get the type
-                        if len(propvalue) == 0:
-                          # just skip for now
-                          #print("Warning: argparse cannot deduce type for property %s of %s. Needs to be set in options file." % (prop, conf.name()))
-                          continue
-                        else:
-                          # deduce type from first item of the list
-                          proptype = type(propvalue[0])
-                        propnargs = "+"
-                        
-                      propName = conf.name() + '.' + prop
-                      # try to add argument once as a shortcut (prop, p.ex '--filename')
-                      # and once with the full property name (propName, p.ex '--PodioOutput.filename')
-                      # if the argument is already present, an ArgumentError is raised that is dealt with below
-                      parser.add_argument("--%s" % prop, "--%s" % propName, type=proptype, help=props[prop][1],
-                        nargs=propnargs,
-                        default=propvalue)
-                      # bookkeeping which option belongs to which configurable
-                      option_db[propName] = (conf, prop)
-                    except argparse.ArgumentError:
-                      # argument already present, add it without shortcut
-                      parser.add_argument( "--%s" % propName, type=proptype, help=props[prop][1],
-                        nargs=propnargs,
-                        default=propvalue)
-                      option_db[propName] = (conf, prop)
-                      pass
+                    if propvalue == conf.propertyNoValue:
+                       propvalue = conf.getDefaultProperty(prop) # thus get the default value
+                    proptype = type(props[prop][0])
+                    # if the property is a list of something, we need to set argparse nargs to '+'
+                    propnargs = "?"
+                    if proptype == list:
+                      # tricky edgecase: if the default is an empty list there is no way to get the type
+                      if len(propvalue) == 0:
+                        # just skip for now
+                        #print("Warning: argparse cannot deduce type for property %s of %s. Needs to be set in options file." % (prop, conf.name()))
+                        continue
+                      else:
+                        # deduce type from first item of the list
+                        proptype = type(propvalue[0])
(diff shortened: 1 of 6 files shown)


k4FWCore/scripts/k4run
```diff
@@ -104,12 +104,20 @@ if __name__ == "__main__":
...
                         default=propvalue)
                       # bookkeeping which option belongs to which configurable
                       option_db[propName] = (conf, prop)
                     except argparse.ArgumentError:
+                      # argument already present, add it without shortcut
+                      parser.add_argument( "--%s" % propName, type=proptype, help=props[prop][1],
+                        nargs=propnargs,
+                        default=propvalue)
+                      option_db[propName] = (conf, prop)
                       pass
```
andresailer, line 121: `pass` isn't needed now

k4FWCore/scripts/k4run
```diff
@@ -104,12 +104,20 @@ if __name__ == "__main__":
                         propnargs = "+"
                         
                       propName = conf.name() + '.' + prop
-                      parser.add_argument("--%s" % propName, type=proptype, help=props[prop][1],
+                      # try to add argument once as a shortcut (prop, p.ex '--filename')
+                      # and once with the full property name (propName, p.ex '--PodioOutput.filename')
+                      # if the argument is already present, an ArgumentError is raised that is dealt with below
+                      parser.add_argument("--%s" % prop, "--%s" % propName, type=proptype, help=props[prop][1],
```
andresailer, line 110: Because of the "completion" argparse does by itself (https://docs.python.org/3/library/argparse.html#allow-abbrev). what about adding 
`nameProp = prop + "." + conf.name()`
Then you can still use --filename if there is no other property with that name and argparse will complete to `--filename.PodioOutput`, but otherwise you get an error, which is probably preferable to setting the wrong option?

Might be better to put the propName first in the list of arguments to have a better visual grouping of the options when doing `--help`

vvolkl: I like that a lot! It feels slightly backwards to begin with prop, but it will simplify things a lot and that's always good.

vvolkl: This won't actually work with `--filename` because of the ambiguity with `--filenameRemote` though. I think we can just rename the properties to reduce the ambiguities, however

andresailer: I meant "adding" as in keeping the "conf.Name().prop" still available. And using that as the "primary" argument for the "visual" grouping of options

```python
confProp = conf.Name() + "." + prop
propConf = prop + "." + conf.Name()
parser.add_argument("--%s" % confProp,
    "--%s" % propConf,
    type=proptype, help=props[prop][1])
```

e.g.:
```
  --InitDD4hep.Blocking [INITDD4HEP.BLOCKING], --Blocking.InitDD4hep [INITDD4HEP.BLOCKING]
                        if algorithm invokes CPU-blocking system calls
                        (offloads computations to accelerators or quantum
                        processors, performs disk or network I/O, is bound by
                        resource synchronization, etc) [Gaudi::Algorithm]
  --InitDD4hep.PrintEmptyCounters [INITDD4HEP.PRINTEMPTYCOUNTERS], --PrintEmptyCounters.InitDD4hep [INITDD4HEP.PRINTEMPTYCOUNTERS]
                        force printing of empty counters, otherwise only
                        printed in DEBUG mode [GaudiCommon<Algorithm>]
  --InitDD4hep.ProcessorType [INITDD4HEP.PROCESSORTYPE], --ProcessorType.InitDD4hep [INITDD4HEP.PROCESSORTYPE]
                        [MarlinProcessorWrapper]
  --InitDD4hep.Parameters INITDD4HEP.PARAMETERS [INITDD4HEP.PARAMETERS ...], --Parameters.InitDD4hep INITDD4HEP.PARAMETERS [INITDD4HEP.PARAMETERS ...]
                        [MarlinProcessorWrapper]
  --VXDBarrelDigitiser.Blocking [VXDBARRELDIGITISER.BLOCKING], --Blocking.VXDBarrelDigitiser [VXDBARRELDIGITISER.BLOCKING]
                        if algorithm invokes CPU-blocking system calls
                        (offloads computations to accelerators or quantum
                        processors, performs disk or network I/O, is bound by
                        resource synchronization, etc) [Gaudi::Algorithm]
  --VXDBarrelDigitiser.PrintEmptyCounters [VXDBARRELDIGITISER.PRINTEMPTYCOUNTERS], --PrintEmptyCounters.VXDBarrelDigitiser [VXDBARRELDIGITISER.PRINTEMPTYCOUNTERS]
                        force printing of empty counters, otherwise only
                        printed in DEBUG mode [GaudiCommon<Algorithm>]
  --VXDBarrelDigitiser.ProcessorType [VXDBARRELDIGITISER.PROCESSORTYPE], --ProcessorType.VXDBarrelDigitiser [VXDBARRELDIGITISER.PROCESSORTYPE]
                        [MarlinProcessorWrapper]
  --VXDBarrelDigitiser.Parameters VXDBARRELDIGITISER.PARAMETERS [VXDBARRELDIGITISER.PARAMETERS ...], --Parameters.VXDBarrelDigitiser VXDBARRELDIGITISER.PARAMETERS [VXDBARRELDIGITISER.PARAMETERS ...]
                        [MarlinProcessorWrapper]
```

k4FWCore/scripts/k4run
```diff
@@ -80,38 +80,36 @@ if __name__ == "__main__":
...
-                      propName = conf.name() + '.' + prop
-                      parser.add_argument("--%s" % propName, type=proptype, help=props[prop][1],
-                        nargs=propnargs,
-                        default=propvalue)
-                      # bookkeeping which option belongs to which configurable
-                      option_db[propName] = (conf, prop)
-                    except argparse.ArgumentError:
-                      pass
-
+                    # argparse ArgumentError: used to catch when trying to add
```
andresailer, line 83: comment no longer valid

PR #52, opened 2021-06-18
Make podio::CollectionBase read from file do prepareForWrite()

### LINKED ISSUE #44: Empty collection when saved by PodioOutput with "keep *"
When attempting to save Podio collection read by `PodioInput` using `PodioOutput` with `keep *` option, it creates an instance of Podio collection in the output file; however, the collection is empty. Hence this breaks links between PODs such as `OneToOneRelations`, throwing segfault when trying to read the output file by `podio::EventStore->get<T>(std::string)` later.

 - OS version: centos7
 - Compiler version: GCC 8.3
 - Package version: v01-00pre09
 - Reproduced by:
```bash
k4run test/k4TestFWCore/options/createExampleEventData.py # creates output_k4test_exampledata.root
k4run test/k4TestFWCore/options/readExampleEventData.py # creates output_k4test_exampledata_2.root
```

#### output_k4test_exampledata.root
cap01

#### output_k4test_exampledata_2.root
cap02

SanghyunKo:
BEGINRELEASENOTES
- Make podio::CollectionBase read from file do prepareForWrite(), fixes #44

ENDRELEASENOTES

Currently, `podio::CollectionBase` read from a file is not saved correctly (though vector members are saved). This seems to be due to missing call of `podio::CollectionBase*->prepareForWrite()`, which is responsible for `m_data->push_back(obj->data)` [[link](https://github.com/AIDASoft/podio/blob/02d613f69dde4f8c571e0084e6036ea89ac048db/python/templates/Collection.cc.jinja2#L114)]. Different handling between collections read from file and written by user is done by a flag `m_isReadFromFile`, and the flag can be set via `podio::CollectionBase*->prepareAfterRead()` after reading the collection from the file [[link](https://github.com/AIDASoft/podio/blob/02d613f69dde4f8c571e0084e6036ea89ac048db/python/templates/Collection.cc.jinja2#L165)]. Then `podio` will take care of it.

Tested with `createExampleEventData.py` and `readExampleEventData.py`, but please let me know if there is something that I overlooked.


vvolkl:
Thanks @SanghyunKo, good catch! Could you still add a test that would have caught this problem, to avoid regression on this issue?

SanghyunKo:
Hi @vvolkl, thanks for the swift response. I added a (sloppy) test that checks `keep` and `drop` switch, just let me know if you think it needs a bit more improvement :)

vvolkl:
That works, thanks!

PR #53, opened 2021-07-13
add get CellID String from DataHandle

fdplacido:
BEGINRELEASENOTES
- Add cellID string recovery from DataHandle through Podio Data Service

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/include/k4FWCore/DataHandle.h
+++ b/k4FWCore/include/k4FWCore/DataHandle.h
@@ -41,6 +41,9 @@ public:
 
   T* put( std::unique_ptr<T> object );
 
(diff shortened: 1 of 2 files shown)


vvolkl (changes requested):
Thanks, it's really about time to refactor things and use the podio EventStore here (should be possible once the user extensions are in).

Could you still add a basic test?

fdplacido:
Yeah I agree, I will add a basic test for this. 
What do you think about getting the cellID directly vs returning the `GenericParameter`?

vvolkl:
I think it's fine -- as said we'll refactor when edm4hep has the UserExtensions. I'll merge now since I'm deploying a new release, you can add a test in a new PR.

PR #56, opened 2021-08-12
use own registerObject when reading collection

fdplacido:
BEGINRELEASENOTES
- After reading with PodioInput, collections are not available even if they were being registered. Wrong `registerObject` was being called.
ENDRELEASENOTES


fdplacido:
I figured out when calling `getCollections()` from PodioDataService this won't return the collections, but `getReadCollections()` will do it. Which I found quite confusing.

Issue #58, opened 2021-08-13
IJobOptionsSvc removed in Gaudi 36.0


tmadlener:
The `IJobOptionsSvc` that is used by the PodioOutput has been deprecated and removed from Gaudi v36 onwards: See [CHANGELOG](https://gitlab.cern.ch/gaudi/Gaudi/-/blob/master/CHANGELOG.md#v36r0-2021-06-21)

We are using it here:
https://github.com/key4hep/k4FWCore/blob/67e5d9452a49f4dee2c3010d0f74f925e4a11b6d/k4FWCore/components/PodioOutput.cpp#L2

It is not entirely clear to me what should be used instead.

mirguest:
Hi Thomas,
Please see my PR #57 .
Tao

tmadlener:
Hi Tao,
I completely missed that. Sorry. But even better if there already is a PR for that ;)
Thomas

PR #59, opened 2021-08-13
Add dev3 and dev4 CI workflows

tmadlener:
BEGINRELEASENOTES
- Add dev3 and dev4 CI workflows

ENDRELEASENOTES

With dev3 we get Gaudis master branch, which might give us a heads-up if things are changing (see, e.g. #58). Maybe make the devX workflows to not be required for PRs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        SETUP: ['/cvmfs/sw.hsf.org/key4hep/setup.sh', '/cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh', '/cvmfs/sft.cern.ch/lcg/views/LCG_99/x86_64-centos7-gcc8-opt/setup.sh']
+        cvmfs_base: ['sft.cern.ch/lcg/views']
+        ENVIRONMENT: ['LCG_99/x86_64-centos7-gcc8-opt',
+                      'dev3/latest/x86_64-centos7-gcc10-opt',
+                      'dev4/latest/x86_64-centos7-clang11-opt']
+        include:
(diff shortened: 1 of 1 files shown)


vvolkl:
Fine by me, but could you separate the cmake configure and build, install and test steps? Make the logs more accessible, and also shows at a glance where a potential problem is.

vvolkl (changes requested):
see comment above

tmadlener:
Removed the `run-lcg-view` action again, as I am not sure it supports workflows with separate steps.

vvolkl:
Great, thanks! I'll go ahead and merge this as there is no harm in having more tests, but using the lcg releases will always have the limitations that we cannot test the `k4*` packages that depend on `k4frwcore` -- at least  if we don't want to start building the packages that are missing in the lcg releases directly in the CI. (This would actually be feasible with some improved support for `spack external` and/or a smart binary cache system)

If we care about building against the HEAD of Gaudi,  the better way is probably to build `gaudi@master` in the nightlies. The only reason this is not currently done, is that I only added the lookup of the commit hash for github, but it definitely makes sense to be able to support gitlab there too.

vvolkl (approved):

vvolkl:
Opened an issue for run-lcg-view: https://github.com/AIDASoft/run-lcg-view/issues/1

tmadlener:
> using the lcg releases will always have the limitations that we cannot test the `k4*` packages that depend on `k4fwcore`

That is actually a good point. My original thinking was that anything that depends on Gaudi in Key4hep in any case depends on FWCore, and that we would catch a lot of problems with using `dev3` and `dev4` here. However, I now realize that there will probably not cover all the Gaudi features already here. Still, we will have at least a few covered.

PR #62, opened 2021-09-14
replace fcc by key4hep in doc strings

hegner:
BEGINRELEASENOTES
- replace fcc by key4hep in doc strings

ENDRELEASENOTES


vvolkl (approved):
Thanks!

PR #63, opened 2021-09-14
remove SimG4 interfaces from core

hegner:
BEGINRELEASENOTES
- Removed SimG4 Interfaces from k4FWCore

ENDRELEASENOTES

I do not think that these interfaces belong into the FWCore package. SimG4 is too specific and IIUC k4SimGeant4 already contains what is needed


vvolkl:
No, k4SimGeant4 does not contain these files anymore, if they are removed here, they need to be added there. In general I think there is some benefit to having all framework interface headers in this package, but if k4SimGeant4 contains all framework code for Geant4, I'm not opposed to also move these interfaces there.

andresailer:
I guess we leave things as they are.

Issue #54, opened 2021-07-16
Add test for CellID retrieval in Datahandle 


vvolkl:
See #53

fdplacido:
I believe this can be closed now.

PR #65, opened 2021-10-12
test getting cellID

fdplacido:
BEGINRELEASENOTES
- Adds test to check getting cellID from DataHandle.
- 2 Algorithms: first writes collection with cellID, second reads it using the DataHandle and checks for correctness.

ENDRELEASENOTES


fdplacido:
Solves https://github.com/key4hep/k4FWCore/issues/54

PR #67, opened 2021-11-02
[WIP] Add test for podiooutput with collections only in first event

fdplacido:
BEGINRELEASENOTES
- This test creates some collections only in specific events, then uses PodioOutput to write the events.
- It will crash on writing the collections.

ENDRELEASENOTES


fdplacido:
Podio assumes all collections will be there at the moment.

jmcarcell:
This is a PR with more than two years and still hasn't been merged, so I'm closing it.

tmadlener:
This will also not be easy to implement with ROOT as an I/O backend, since that assumes the same structure for every event.

PR #74, opened 2022-01-28
Fix usage of unique_ptr from upstream podio

tmadlener:
BEGINRELEASENOTES
- Fixing small problem due to AIDASoft/podio#252 introducing the usage of `std::unique_ptr` in some cases for the collection buffers.

ENDRELEASENOTES


vvolkl (approved):

tmadlener:
Now also fixing a runtime issue discovered by @fdplacido

PR #80, opened 2022-04-01
Add Gaudi Service to generate Unique IDs based on event number, run number and Gaudi Algorithm

fdplacido:
BEGINRELEASENOTES
- Add Gaudi Service to generate Unique IDs based on event number, run number and Gaudi Algorithm

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/CMakeLists.txt
+++ b/k4FWCore/CMakeLists.txt
@@ -28,7 +28,18 @@ target_include_directories(k4FWCorePlugins PUBLIC
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 
 
-install(TARGETS k4FWCore k4FWCorePlugins
+gaudi_add_module(UniqueIDGenSvc
+    SOURCES
+      src/UniqueIDGenSvc.cpp
+    LINK
+      Gaudi::GaudiAlgLib
+)
+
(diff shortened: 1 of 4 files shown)


k4FWCore/CMakeLists.txt
```diff
@@ -28,7 +28,18 @@ target_include_directories(k4FWCorePlugins PUBLIC
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 
 
-install(TARGETS k4FWCore k4FWCorePlugins
+gaudi_add_module(UniqueIDGenSvc
```
vvolkl, line 31: Why not add this service to the k4FWCorePlugins module?

vvolkl:
Looks good to me, but before merging it, could you add a basic test?

vvolkl (changes requested):
Thanks for adding the test! I'd still prefer building the `UniqueIDGenSvc` component as part of the  `k4FWCorePlugins` module. Adding a new module just seems to add overhead, or is there a reason I'm not seeing?

fdplacido:
> Thanks for adding the test! I'd still prefer building the `UniqueIDGenSvc` component as part of the `k4FWCorePlugins` module. Adding a new module just seems to add overhead, or is there a reason I'm not seeing?

I kept it separated because we are not installing the headers of the components, so to not mix it. But I can move things there and install it for this case.

Issue #81, opened 2022-04-22
Add Documentation for UniqueIDGenSvc


andresailer:
Somewhere, probably here

https://github.com/key4hep/k4FWCore/blob/d6555ad5679fb10f194636a52b0bd5eee272749c/k4FWCore/components/UniqueIDGenSvc.h#L3

There should be an explanation what this service should be used for. And contain all the relevant keywords: `reproducible, random, seed`

fdplacido:
I agree, I will add some comments to the header for that.
There is a doc md file here with documentation: https://github.com/fdplacido/k4MarlinWrapper/blob/random_gen_doc/doc/randomNumberGeneration.md
It is not yet merged in either project

PR #83, opened 2022-04-25
Add comments for uniqueID svc

fdplacido:
BEGINRELEASENOTES
- Add Unique ID generation service
- Add test for UniqueIDGenSvc

ENDRELEASENOTES

Solves #81

### DIFF
--- a/k4FWCore/components/UniqueIDGenSvc.cpp
+++ b/k4FWCore/components/UniqueIDGenSvc.cpp
@@ -15,9 +15,9 @@ const size_t bits64 = std::numeric_limits<uint64_t>::digits;
 const size_t bitsSizeT = std::numeric_limits<size_t>::digits;
 
 size_t UniqueIDGenSvc::getUniqueID(
-  int64_t seed, uint32_t evt_num, uint32_t run_num, std::string name) const
+  uint32_t evt_num, uint32_t run_num, std::string name) const
 {
-  std::bitset<bits64> seed_bits(seed);
+  std::bitset<bits64> seed_bits(this->m_seed);
   std::bitset<bits32> event_num_bits(evt_num), run_num_bits(run_num);
   size_t str_hash = std::hash<std::string>{}(name);
   std::bitset<bitsSizeT> name_bits(str_hash);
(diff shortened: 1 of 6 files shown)


fdplacido:
Solves https://github.com/key4hep/k4FWCore/pull/67

vvolkl:
Can you use doxygen-style comments? There is no doxygen code reference built from this at the moment, but maybe we want one in the future and then it would be nice to have this consistent. Example: https://github.com/key4hep/k4-project-template/blob/master/k4ProjectTemplate/src/components/CreateExampleEventData.h#L18

k4FWCore/components/UniqueIDGenSvc.h
```diff
@@ -7,6 +7,10 @@
 #include <GaudiKernel/Service.h>
 #include "k4Interface/IUniqueIDGenSvc.h"
 
+
+// Generate a random, unique*, reproducible number using
+// event number, event run, and algorithm name.
+// To be used as ID across events and algorithms.
```
andresailer:
```diff
-// Generate a random, unique*, reproducible number using
-// event number, event run, and algorithm name.
-// To be used as ID across events and algorithms.
+// Generate a unique, reproducible number using
+// event number, run number, algorithm name, and global seed
+// To be used as reproducible per event seed for random number generators used inside algorithms.
```
Actually that "global seed" is not yet implemented

andresailer: I.e. the service should have a seed property that allows one to globally change all the specific seeds, but same global  seed results in otherwise same unique seeds.

fdplacido:
This test will fail until latest Gaudi is in the nightlies. See: https://gitlab.cern.ch/gaudi/Gaudi/-/merge_requests/1321

k4FWCore/components/UniqueIDGenSvc.cpp
```diff
@@ -11,18 +11,22 @@ StatusCode UniqueIDGenSvc::initialize() {
...
+const size_t bits64 = std::numeric_limits<uint64_t>::digits;
 const size_t bitsSizeT = std::numeric_limits<size_t>::digits;
 
 size_t UniqueIDGenSvc::getUniqueID(
-  uint32_t evt_num, uint32_t run_num, std::string name) const 
+  int64_t seed, uint32_t evt_num, uint32_t run_num, std::string name) const
 {
-  std::bitset<bits32> event_num_bits(evt_num);
-  std::bitset<bits32> run_num_bits(run_num);
+  std::bitset<bits64> seed_bits(seed);
```
andresailer, line 20: This has to be like this
```suggestion
  std::bitset<bits64> seed_bits(this->seed);
```
And the service should have a seed property, and the function signature should not change.

PR #91, opened 2022-12-10
Fix a deprecation warning

jmcarcell:
BEGINRELEASENOTES

- Use the new templated getters, fixes compiler warnings

ENDRELEASENOTES

The non-templated getters were deprecated in https://github.com/AIDASoft/podio/pull/262

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,8 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         cvmfs_base: ['sft.cern.ch/lcg/views']
-        ENVIRONMENT: ['LCG_102rc1/x86_64-centos7-gcc11-opt',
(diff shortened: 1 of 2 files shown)


vvolkl:
Thanks! This introduces a new dependency version constraint as it needs podio v00-14-02 or newer. This is not currently available in the lcg releases, so you'd need to remove LCG_102rc1 here: https://github.com/key4hep/k4FWCore/blob/master/.github/workflows/test.yml#L12

I think we can keep the dev3 workflow even though it is currently broken, seems like a temporary publication failure only, in principle it should work.

Ideally you could also update the podio dependency in the recipe of our build tool to reflect this change: https://github.com/key4hep/key4hep-spack/blob/release/packages/k4fwcore/package.py#L26

andresailer:
There are no more gcc10 dev3/dev4 builds we have to move to gcc11 for that.

Issue #90, opened 2022-11-28
Do not attach gdb by default


tmadlener:
Currently `k4run` automatically attaches and runs gdb on a crash to get a backtrace. To arrive at an actual backtrace a significant amount of time passes. I think this should not be the default.

tmadlener:
I was wrong with attributing the fault to `k4run`. It is actually root that attaches the debugger on a crash. Hence, to cut out the debugger it is necessary to change the `Root.Stacktrace` setting to `no` in an appropriate `.rootrc` file, as the default one that is shipped with root in `$ROOTSYS/etc/system.rootrc` contains
```
Root.Stacktrace:    yes
```

A `~/.rootrc` with
```
Root.Stacktrace: no
```
will disable the stacktrace "globally", however, a `./.rootrc` in the folder where `k4run` is executed with the same contents offers a more granular disabling of this.

Issue #89, opened 2022-11-28
k4DataSvc seg faults if input file does not exist


tmadlener:
When trying to open a non existing file, `k4run` will seg fault. This should be handled much more gracefully, as typos are otherwise pretty time consuming

tmadlener:
This is potentially fixed with AIDASoft/podio#95. Need to check again

jmcarcell:
This doesn't seem to happen anymore:
``` 
TFile::TFile        ERROR file /path/to/wrong_file.root does not exist
EventDataSvc        FATAL in sysInitialize(): standard std::exception is caught
```
Feel free to reopen if it seg faults again

PR #93, opened 2023-01-12
Improvements to k4run

jmcarcell:
BEGINRELEASENOTES
- Add improvements to k4run: clean up code, use the logging module, parse all the arguments using argparse, sort display of options

ENDRELEASENOTES

- Parsing is done using `argparse` only instead of the previous combination of manual parsing and `argparse`. I think this is cleaner and if the arguments are wrong it will show the usage automatically (before there was a explicit testing of this)
- Code was cleaned up a bit, for example there were very deeply nested if/for that can be changed a bit. Indentation is (and was) inconsistent so if we agree on 2 or 4 spaces I can change that; it makes the diff very hard to read so we can keep it for the last change.
- Previously the display of options wouldn't be sorted and would be different each time because there was a `list(set(` that has undetermined ordering, now it will always be sorted in the same order as they were defined in the file (for python >= 3.7, since it is a one liner in that case).
- Now using `logging` for messages as per #50. Right now there is a `[k4run]` for the messages that come from `k4run` but this is a placeholder. For example for the Gaudi processes the format looks like `[ MESSAGE "VXDEndcapDigitiser"]` and many things can be done like adding the date, the level of the message, etc. I don't have a strong opinion. This is how it looks now:

``` python
[k4run] LcioEvent --> MyAIDAProcessor --> EventNumber --> InitDD4hep --> Config --> OverlayFalse --> VXDBarrelDigitiser --> VXDEndcapDigitiser --> InnerPlanarDigiProcessor --> InnerEndcapPlanarDigiProcessor --> OuterPlanarDigiProcessor --> OuterEndcapPlanarDigiProcessor --> MyConformalTracking --> ClonesAndSplitTracksFinder --> Refit --> MyDDCaloDigi --> MyDDSimpleMuonDigi --> MyDDMarlinPandora --> LumiCalReco --> MergeRP --> MergeClusters --> MyClicEfficiencyCalculator --> MyRecoMCTruthLinker --> MyTrackChecker --> CLICPfoSelectorDefault_HE --> CLICPfoSelectorLoose_HE --> CLICPfoSelectorTight_HE --> CLICPfoSelectorDefault_LE --> CLICPfoSelectorLoose_LE --> CLICPfoSelectorTight_LE --> RenameCollection --> VertexFinder --> JetClusteringAndRefiner --> Output_REC --> Output_DST
[k4run] Option name: EventDataSvc.OutputLevel EventDataSvc.OutputLevel 0
[k4run] Option name: LcioEvent.OutputLevel LcioEvent.OutputLevel 4
[k4run] Option name: LcioEvent.Files LcioEvent.Files ['ttbar.slcio']
[k4run] Option name: InitDD4hep.OutputLevel InitDD4hep.OutputLevel 4
[k4run] Option name: InitDD4hep.ProcessorType InitDD4hep.ProcessorType InitializeDD4hep
```

I ran `clicReconstruction.py` on the same file and got similar results 

Other related issues:
- #89 I could reproduce with the podio reader but this has been fixed in https://github.com/AIDASoft/podio/pull/355. With the LCIO reader it exits without crashing. Can we close it?
- #90 <strike>I checked and certainly by default the option to attach `gdb` is set to false for Gaudi. I could reproduce this issue with the podio reader before https://github.com/AIDASoft/podio/pull/355 (by trying to open a non-existent file, now fixed), where `TFile.Open` fails and pulls `gdb`. It was very very slow. Interestingly I ran with everything local (without `cvmfs`) and it took just a couple of seconds to display the backtrace, so `cvmfs` is at least partly responsible for the slowness.</strike> Already closed.
- #72 Using `--npcus` with a number different from 0 fails with:

``` python
  File "/cvmfs/sw.hsf.org/spackages6/gaudi/36.5/x86_64-centos7-gcc11.2.0-opt/z2prd/python/Gaudi/Main.py", line 476, in run
    result = self.runParallel(ncpus)
  File "/cvmfs/sw.hsf.org/spackages6/gaudi/36.5/x86_64-centos7-gcc11.2.0-opt/z2prd/python/Gaudi/Main.py", line 561, in runParallel
    Parall = gpp.Coord(ncpus, c, self.log)
  File "/cvmfs/sw.hsf.org/spackages6/gaudi/36.5/x86_64-centos7-gcc11.2.0-opt/z2prd/python/GaudiMP/GMPBase.py", line 1378, in __init__
    sub = Subworker(
  File "/cvmfs/sw.hsf.org/spackages6/gaudi/36.5/x86_64-centos7-gcc11.2.0-opt/z2prd/python/GaudiMP/GMPBase.py", line 834, in __init__
    GMPComponent.__init__(
  File "/cvmfs/sw.hsf.org/spackages6/gaudi/36.5/x86_64-centos7-gcc11.2.0-opt/z2prd/python/GaudiMP/GMPBase.py", line 473, in __init__
    self.msgFormat = self.config["MessageSvc"].Format
KeyError: 'MessageSvc'
```

### DIFF
--- a/k4FWCore/scripts/k4run
+++ b/k4FWCore/scripts/k4run
@@ -76,13 +76,11 @@ if __name__ == "__main__":
     # ensure that we (and the subprocesses) use the C standard localization
     os.environ['LC_ALL'] = 'C'
 
-    logger = logging.getLogger(__name__)
+    logger = logging.getLogger()
     logger.setLevel(logging.INFO)
-    # Don't propagate to the root logger, otherwise the output will be doubled
-    logger.propagate = 0
     # formatter = logging.Formatter('[%(asctime)s %(levelname)s] %(message)s', datefmt='%Y-%b-%d %H:%M:%S')
     formatter = logging.Formatter('[k4run] %(message)s')
-    handler = logging.StreamHandler()
+    handler = logging.StreamHandler(stream=sys.stdout)
     handler.setFormatter(formatter)
     logger.addHandler(handler)
 
@@ -116,7 +114,7 @@ if __name__ == "__main__":
     opts = parser.parse_args()
 
     # print a doc line showing the configured algorithms
-    logger.info('-->' + ' --> '.join([alg.name() for alg in ApplicationMgr().TopAlg]))
+    logger.info('--> ' + ' --> '.join([alg.name() for alg in ApplicationMgr().TopAlg]))
 
     if opts.list:
         from Gaudi import Configuration
@@ -155,14 +153,6 @@ if __name__ == "__main__":
         elif opts.ncpus < -1:
             s = "Invalid value : --ncpus must be integer >= -1"
             parser.error(s)
-
-    # configure the logging
-    from GaudiKernel.ProcessJobOptions import InstallRootLoggingHandler
-
-    prefix = "# "
-    level = logging.INFO
-    InstallRootLoggingHandler(prefix, level=level, with_time=opts.verbose)
-    root_logger = logging.getLogger()
 
     from Gaudi.Main import gaudimain
     c = gaudimain()
--- a/k4FWCore/scripts/k4run
+++ b/k4FWCore/scripts/k4run
@@ -116,12 +116,12 @@ if __name__ == "__main__":
     opts = parser.parse_args()
 
     # print a doc line showing the configured algorithms
-    logger.info(' --> '.join([alg.name() for alg in ApplicationMgr().TopAlg]))
+    logger.info('-->' + ' --> '.join([alg.name() for alg in ApplicationMgr().TopAlg]))
 
     if opts.list:
         from Gaudi import Configuration
         cfgDb = Configuration.cfgDb
-        logger.info("Available components:\n%s" % (21 * "="))
+        logger.info("Available components:\n%s", (21 * "="))
         for item in sorted(cfgDb):
             if True:  # another option could filter Gaudi components here
               try:
@@ -136,7 +136,7 @@ if __name__ == "__main__":
 
     opts_dict = vars(opts)
     for optionName, propTuple in option_db.items():
-      logger.info("Option name: " + str(propTuple[1]) + " " + optionName + " " + str(opts_dict[optionName]))
+      logger.info("Option name: %s %s %s", propTuple[1], optionName, opts_dict[optionName])
       propTuple[0].setProp(propTuple[1].rsplit(".",1)[1], opts_dict[optionName])
 
     if opts.verbose:
--- a/k4FWCore/scripts/k4run
+++ b/k4FWCore/scripts/k4run
@@ -4,6 +4,8 @@ from __future__ import print_function
 import os
 import sys
 import argparse
+from multiprocessing import cpu_count
+import logging
 
 
 # these default properties are filtered as otherwise they will clutter the argument list
@@ -55,10 +57,9 @@ if __name__ == "__main__":
     #   -1   : All available cpus
     #    0   : Serial Mode (traditional gaudirun)
     #    n>0 : parallel with n cpus (n <= sys_cpus)
-    parser.add_argument("--ncpus",  type=int, default=0,
+    parser.add_argument("--ncpus", type=int, default=0,
                       help="Start Gaudi in parallel mode using NCPUS processes. "
                            "0 => serial mode (default), -1 => use all CPUs")
-
 
 
     # print a doc line showing the configured algorithms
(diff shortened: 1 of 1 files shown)


k4FWCore/scripts/k4run
```diff
@@ -60,14 +60,9 @@ class LoadFromFile(argparse.Action):
                 proptype = type(propvalue[0])
               propnargs = "+"
 
-          # add the argument twice, once as "--PodioOutput.filename"
-          # and once as "--filename.PodioOutput"
           propName = conf.name() + '.' + prop
           propNameReversed = prop + '.' + conf.name()
-          option_db[propName] = (conf, propName)
-          parser.add_argument( "--%s" % propName, "--%s" % propNameReversed, type=proptype, help=props[prop][1],
```
vvolkl: Why do you remove this? This was added on purpose: Argparse autocompletes if possible, so adding `--filename.PodioOutput` allows using shorthand arguments like `--filename` (at least if filename is a unique property in the configurables

vvolkl:
See my comment on the commit: https://github.com/key4hep/k4FWCore/pull/93/commits/59b0c68f60d131bcd56fa28b079a8a42291e68cb#r1068103342

The 
```
[k4run] Option name: EventDataSvc.OutputLevel EventDataSvc.OutputLevel 0
[k4run] Option name: LcioEvent.OutputLevel LcioEvent.OutputLevel 4
[k4run] Option name: LcioEvent.Files LcioEvent.Files ['ttbar.slcio']
[k4run] Option name: InitDD4hep.OutputLevel InitDD4hep.OutputLevel 4
[k4run] Option name: InitDD4hep.ProcessorType InitDD4hep.ProcessorType InitializeDD4hep
```
logging was added for some ad-hoc debugging, in principle this information is stored in the metadata by  the joboptions service. It could be removed, but maybe it is nice to have in the logs.

k4FWCore/scripts/k4run
```diff
@@ -18,35 +20,83 @@ __filterGaudiProps = [ "ContextService", "Cardinality", "Context", "CounterList"
...
-      if f[0] != '-':
-        exec(open(f).read())
-        run_with_options_file = True
-      else:
-        break
+    os.environ['LC_ALL'] = 'C'
+
+    logger = logging.getLogger(__name__)
+    logger.setLevel(logging.INFO)
+    # Don't propagate to the root logger, otherwise the output will be doubled
```
andresailer, line 81: At what point is the output doubled? If there is already output that means the logger is already created somewhere else, and we don't have to do it again, right?

andresailer: If I remember correctly, if not changed the logging output goes to stderr (and not stdout), where does the output go to at the moment?

k4FWCore/scripts/k4run
```diff
@@ -55,68 +105,25 @@ if __name__ == "__main__":
...
-                    # and once as "--filename.PodioOutput"
-                    propName = conf.name() + '.' + prop
-                    propNameReversed = prop + '.' + conf.name()
-                    option_db[propName] = (conf, propName)
-                    parser.add_argument( "--%s" % propName, "--%s" % propNameReversed, type=proptype, help=props[prop][1],
-                      nargs=propnargs,
-                      default=propvalue)
-
-    opts=parser.parse_args()
+    logger.info(' --> '.join([alg.name() for alg in ApplicationMgr().TopAlg]))
```
andresailer, line 119: This will have the first algorithm not be preceded by `-->`

k4FWCore/scripts/k4run
```diff
@@ -127,13 +134,9 @@ if __name__ == "__main__":
...
         sys.exit()
 
-    if not run_with_options_file:
-      print("usage: " + os.path.basename(__file__) + " [gaudi_config.py ...] Specify a gaudi options file to run")
-      sys.exit()
-
     opts_dict = vars(opts)
     for optionName, propTuple in option_db.items():
-      print("Option name:", propTuple[1], optionName, opts_dict[optionName])
+      logger.info("Option name: " + str(propTuple[1]) + " " + optionName + " " + str(opts_dict[optionName]))
```
andresailer, line 139:
```suggestion
      logger.info("Option name: %s %s %s", propTuple[1], optionName, opts_dict[optionName])
```

k4FWCore/scripts/k4run
```diff
@@ -155,9 +157,7 @@ if __name__ == "__main__":
             parser.error(s)
 
     # configure the logging
-    import logging
-    from GaudiKernel.ProcessJobOptions import (InstallRootLoggingHandler,
-                                               PrintOff)
+    from GaudiKernel.ProcessJobOptions import InstallRootLoggingHandler
```
andresailer, line 160: Seems that should be better merged with the logging config done above now

k4FWCore/scripts/k4run
andresailer: Under what circumstances did you test things when removing the InstallRootLoggingHandler?
What happens if you move that call here instead of the logging setup you are adding?

PR #96, opened 2023-02-25
Fix warnings due to missing override specification

gganis:
BEGINRELEASENOTES
- Fix warnings due to missing override specification

ENDRELEASENOTES


tmadlener (approved):

tmadlener:
Are these the only places where this gets flagged? Naively I would expect there to be more, since we are probably implementing more virtual functions, but maybe I am thinking about test code(?).

tmadlener:
I think only using either `override` or `final` and dropping `virtual` makes both compilers happy, at least it does for podio.

jmcarcell:
Closing this one in favor of #143

PR #97, opened 2023-02-25
Make sure libGaudiCoreSvc.so is found by genconf

gganis:
Fixes several dlopen failures on MacOsX of this sort:

WARNING: cannot load libGaudiCoreSvc.so for factory ApplicationMgr
WARNING: dlopen(libGaudiCoreSvc.so, 0x0009): tried: 'libGaudiCoreSvc.so' (no such file), '/usr/local/lib/libGaudiCoreSvc.so' (no such file), '/usr/lib/libGaudiCoreSvc.so' (no such file), '/Users/ganis/local/key4hep/build/k4FWcore/test/k4FWCoreTest/libGaudiCoreSvc.so' (no such file)

Setting LD_LIBRARY_PATH and similar does not work, but perhaps there is another way to fix it.

BEGINRELEASENOTES
- Fixes several dlopen failures on MacOsX due to libGaudiCoreSvc.so not being found.
ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,8 @@ find_package(ROOT COMPONENTS RIO Tree)
 #---------------------------------------------------------------
 # Load macros and functions for Gaudi-based projects
 find_package(Gaudi)
+get_filename_component(Gaudi_CMakeDir ${GAUDI_TOOLBOX_DIR} DIRECTORY)
+get_filename_component(GAUDI_LIB_DIR ${Gaudi_CMakeDir} DIRECTORY)
 #---------------------------------------------------------------
 
 include(GNUInstallDirs)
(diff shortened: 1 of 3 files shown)


vvolkl:
Hi Gerri, yes, I'm hoping this issue will be fixed upstream, see issue https://gitlab.cern.ch/gaudi/Gaudi/-/issues/251 . 
Using the current working directory for the libraries is possible, but it has the downside that you would need to do this for all gaudi modules, so most key4hep packages as well, which seems cumbersome. https://github.com/key4hep/k4FWCore/pull/98 is an alternative fix that should work for everything, provided you are not using the system python, but an installation outside of /usr/bin.

jmcarcell:
The original issue in Gaudi that Valentin linked has been closed as it is fixed, so I'm closing this one. I did not try to reproduce, however.

PR #98, opened 2023-02-26
MacOS: fix for dlopen errors due to SIP

vvolkl:


k4FWCore/CMakeLists.txt
```diff
@@ -4,6 +4,12 @@
 
 find_package(podio)
 
+if(APPLE)
+  file(READ scripts/k4run filedata)
+  string(REGEX REPLACE "#!/usr/bin/env python" "#!${Python_EXECUTABLE}" filedata "${filedata}")
+  file(WRITE scripts/k4run "${filedata}")
```
andresailer, line 10: This is going to mess with git and the inclusion of this file k4run file.
Better use a k4run.in and configure to set the python executable?

andresailer:
outdated

PR #99, opened 2023-02-28
Setting RPATH

kjvbrt:
BEGINRELEASENOTES
- Sets RPATH, so installed libs are not stripped
ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,9 @@ if(NOT CMAKE_CXX_STANDARD MATCHES "14|17")
   message(FATAL_ERROR "Unsupported C++ standard: ${CMAKE_CXX_STANDARD}")
 endif()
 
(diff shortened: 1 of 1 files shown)


andresailer:
Please use the whole RPATH things as linked to from  https://github.com/key4hep/k4-project-template/issues/12

kjvbrt:
Would `K4FWCORE_SET_RPATH` option be useful?

andresailer:
Thanks. Probably the option will be useful?

kjvbrt:
I mean, whether there is a need to not set RPATH for the installed executables/libs.

andresailer:
If you put everything in a "view" structure in the end, you don't want RPATH, so it could be useful.

jmcarcell:
Let's merge this which would fix one issue. We should also have this in `k4-project-template`.

Issue #95, opened 2023-02-23
PodioOutput saves metadata of droped collections into the ROOT file


kjvbrt:
When collection is dropped in PodioOutput its metadata are saved regardless.

 - OS version: e.g. centos7
 - Package version: key4hep-nightlies
 - Reproduced by: `podio-dump` on generated file crashes when dropped collection is encountered
 - Example file generator: https://github.com/HEP-FCC/k4RecCalorimeter/blob/main/RecFCCeeCalorimeter/tests/options/runCaloSim.py
 - Expected behaviour: Metadata of dropped collections are not saved

tmadlener:
Could you also post the backtrace of the `podio-dump` crash? Have you also tried what happens if you don't drop any collections at all in the podio output?

tmadlener:
I can reproduce this. I am not yet entirely sure if this is an issue in the `ROOTLegacyReader` that is invoked for reading the produced files or if this is truly an issue of the `PodioOutput` algorithm.

kjvbrt:
> Could you also post the backtrace of the `podio-dump` crash? Have you also tried what happens if you don't drop any collections at all in the podio output?

[crash.txt](https://github.com/key4hep/k4FWCore/files/10841632/crash.txt)

If none of the collections is dropped, then there is no crash.

kjvbrt:
Reading the files using `ROOTReader` produces following errors, but does not crash.
```
PODIO: Reconstructing CollectionTypeInfo branch from other sources in file: 'output_fullCalo_SimAndDigi.root'
Problems reconstructing collection info for collection: 'ECalBarrelPositionedHits'
Problems reconstructing collection info for collection: 'ECalEndcapPositionedHits'
Problems reconstructing collection info for collection: 'HCalBarrelPositionedHits'
Problems reconstructing collection info for collection: 'ECalBarrelCellsStep1'
Problems reconstructing collection info for collection: 'ECalBarrelCellsStep2'
Problems reconstructing collection info for collection: 'HCalBarrelCells'
Problems reconstructing collection info for collection: 'emptyCaloCells'
Problems reconstructing collection info for collection: 'CaloClusterCells'
```

kjvbrt:
[...]

PR #103, opened 2023-05-17
[WIP] Add automatic Frame legacy read

hegner:
BEGINRELEASENOTES
- Transparent handling when reading legacy PODIO files

ENDRELEASENOTES

This PR is WIP. It seems to suffer from a bug in underlying libs which need to be addressed first.

### DIFF
--- a/k4FWCore/components/PodioLegacyOutput.cpp
+++ b/k4FWCore/components/PodioLegacyOutput.cpp
@@ -122,7 +122,7 @@ void PodioLegacyOutput::createBranches(const std::vector<std::pair<std::string, 
     const auto collID = m_podioLegacyDataSvc->getCollectionIDs()->collectionID(collName).value();
     // No check necessary, only registered collections possible
     auto       coll     = collNamePair.second;
-    const auto collType = std::string(coll->getValueTypeName()) + "Collection";
+    const auto collType = coll->getTypeName();
     collectionInfo->emplace_back(collID, std::move(collType), coll->isSubsetCollection());
     //}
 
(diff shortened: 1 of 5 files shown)


hegner:
@tmadlener - working on this showed an interesting incident, why moving to frames was the right thing. We have https://github.com/key4hep/k4FWCore/blob/master/k4FWCore/components/PodioOutput.cpp#L108 that uses the old schema for the collection info. Even though the PODIO used does the newer schema. Thus legacy files written with PodioOutput are not compatible with plain PODIO. Unfortunately we may need to put the error handling into PODIO itself into the ROOTLegacyReader.

tmadlener:
Could the old schema version for collection info be the root cause for #95?

Putting the handling of that into the ROOTLegacyReader sounds fine to me.

hegner:
Yes, that could be an explanation for #95.

andresailer:
This needs a rebase and conflict resolution if still valid?

tmadlener:
Rebased onto the latest master and things should be compiling again. As discussed during the last key4hep meeting, this should be able to read legacy files as written by the `podio::ROOTWriter`, but will most likely fail to read legacy files produced by k4FWCore. In order to make that work it is still necessary to repackage the current (k4FWCore) legacy input system to produce Frame(Data).

jmcarcell:
Can we close this? We are already going to have files pre EDM4hep 1.0 (these will be the "legacy" ones) and after EDM4hep 1.0

tmadlener:
I think we can close this for now and come back if we actually need this functionality in the future.

Issue #109, opened 2023-06-15
`k4run` crashes when invoked without arguments


tmadlener:
Originally reported here: https://github.com/key4hep/key4hep-spack/issues/498

```
* ********************************************************************
* Welcome to lxplus924.cern.ch, AlmaLinux release 9.2 (Turquoise Kodkod)
* Archive of news is available in /etc/motd-archive
* Reminder: you have agreed to the CERN
*   computing rules, in particular OC5. CERN implements
*   the measures necessary to ensure compliance.
*   https://cern.ch/ComputingRules
* Puppet environment: production, Roger state: production
* Foreman hostgroup: lxplus/nodes/login
* Availability zone: cern-geneva-b
* LXPLUS Public Login Service - http://lxplusdoc.web.cern.ch/
* An AlmaLinux8 based lxplus8.cern.ch is now available
* Please read LXPLUS Privacy Notice in http://cern.ch/go/TpV7
* ********************************************************************
Last login: Thu Jun 15 14:35:46 2023 from 2001:1458:202:96::101:4142
~$ source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh 
AlmaLinux 9 detected
Setting up the latest Key4HEP software stack from CVMFS ...
 ...  Key4HEP release: 2023-06-15
 ...  Use the following command to reproduce the current environment: 
 ... 
         source /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2023-06-15/x86_64-almalinux9-gcc11.3.1-opt/key4hep-stack/2023-06-15-zkepcb/setup.sh
 ... 
 ...  If you have any issues, comments or requests open an issue at https://github.com/key4hep/key4hep-spack/issues
~$ k4run
Traceback (most recent call last):
  File "/cvmfs/sw-nightlies.hsf.org/key4hep/releases/2023-06-15/x86_64-almalinux9-gcc11.3.1-opt/k4fwcore/348be068b687353f2886421110190aaa682bd293=develop-yyw2xj/bin/k4run", line 113, in <module>
    opts = parser.parse_known_args()
  File "/cvmfs/sw-nightlies.hsf.org/key4hep/releases/2023-06-10/x86_64-almalinux9-gcc11.3.1-opt/python/3.10.10-mbuvpt/lib/python3.10/argparse.py", line 1859, in parse_known_args
    namespace, args = self._parse_known_args(args, namespace)
[...]

PR #110, opened 2023-06-15
Fix crash without arguments

jmcarcell:
Fixes https://github.com/key4hep/k4FWCore/issues/109
and now exits unsuccessfully


tmadlener:
That was quick :) Thanks. Should we add a simple test for this? Since the reproducer is almost trivial this should be straight forwardish(?)

It seems like I have missed one of the things related to key4hep/EDM4hep#188 which causes the nightlies based CI to fail.

jmcarcell:
Yeah I can add a test. This shouldn't be an EDM4hep because it's the same version that has been built for the nightlies (which built fine today)

jmcarcell:
Actually it was: https://github.com/key4hep/k4FWCore/pull/111
Why didn't it break the nightlies? Hmm...

tmadlener:
do you build / run the tests in the nightlies? Because here the problem is in there

test/k4FWCoreTest/CMakeLists.txt
```diff
@@ -107,3 +107,8 @@ add_test(NAME TestUniqueIDGenSvc
          WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
          COMMAND ${K4RUN} options/TestUniqueIDGenSvc.py)
 set_test_env(TestUniqueIDGenSvc)
+
+add_test(NAME Testk4runNoArguments
+         WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
+         COMMAND ${K4RUN})
+set_test_env(Testk4runNoArguments)
```
tmadlener, line 114:
```suggestion
set_test_env(Testk4runNoArguments)
set_tests_properties(Testk4runNoArguments
  PROPERTIES PASS_REGULAR_EXPRESSION "Usage: k4run <options_file.py>, use --help to get a complete list of arguments")
```

tmadlener (approved):

PR #119, opened 2023-07-06
Introduce a quick workaround for collection metadata creation

tmadlener:
BEGINRELEASENOTES
- Make (mutable) access to metadata Frame public temporarily to allow for a timely solution of some of the issues in the marlin wrapper.
- Remove function for creating collection parameter keys in favor of the one introduce in upstream podio in AIDASoft/podio#439

ENDRELEASENOTES


tmadlener:
As discussed during todays meeting we will also merge this in its current form, once it builds again, in order to unblock some downstream changes and properly fix it after the "summer break".

PR #124, opened 2023-07-18
Make MetaDataHandle handle all keys correctly

tmadlener:
BEGINRELEASENOTES
- Make the `MetaDataHandle` work corretly with `DataHandles` that do not have `"/Event/"` in their descriptor
- Mark `get` and `fullDescriptor` as `const` as they don't mutate any internal state

ENDRELEASENOTES


tmadlener:
Merging this as it is in order to let the wrapper PR move forward

PR #129, opened 2023-07-27
Use Gaudi::Functional, add examples and adapt PodioInput to use Gaudi::Functional

jmcarcell:
BEGINRELEASENOTES
- Make changes in `DataWrapper.h` so that algorithms using Gaudi::Functional can work
- Add examples and tests of a consumer, producer and transformer and their versions with multiple inputs and/or outputs
- Adapt `PodioInput` to use Gaudi::Functional

ENDRELEASENOTES

This PR keeps the old tests (that still work) that we can remove or adapt later. There should be no breaking changes and so far the changes to `PodioInput` seem to work fine.


test/k4FWCoreTest/CMakeLists.txt
```diff
@@ -136,4 +142,40 @@ add_test(NAME Testk4runNoArguments
...
+
+add_test(NAME ExampleFunctionalConsumerMultiple
+         WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
+         COMMAND ${K4RUN} options/runExampleFunctionalProducerMultiple.py)
+set_test_env(ExampleFunctionalConsumerMultiple)
+set_tests_properties(ExampleFunctionalConsumerMultiple PROPERTIES DEPENDS ExampleFunctionalProducerMultiple)
+
+add_test(NAME ExampleFunctionalTransformerMultiple
+         WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
+         COMMAND ${K4RUN} options/runExampleFunctionalTransformerMultiple.py)
```
Zehvogel, line 179: Did you forget to push this file?

Zehvogel:
Nice that the tests work now! :)

I am still a bit confused about the semantics, specifically using a consumer for reading.
My Impression was that this would be done by a producer in the functional approach, but this is completely side-stepped by the DataSvc? PodioInput now behaves like a constant (if there is such a thing in Gaudi functional) as you are not taking any inputs and don't produce any outputs (directly).

test/k4FWCoreTest/src/components/ExampleFunctionalConsumer.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include <string>
+
+// This will always be Gaudi::Algorithm
+using BaseClass_t = Gaudi::Functional::Traits::BaseClass_t<Gaudi::Algorithm>;
+
+// Which type of collection we are reading
+// this will always be podio::CollectionBase
+// Has to be wrapped in DataWrapper
+using colltype = DataWrapper<podio::CollectionBase>;
```
hegner, line 17: As just discussed in person you can avoid these using statements and the user side visibility of handles and wrappers by changing BaseClass_t.
We are using our special wrappers because of what is needed for PODIO [here](https://github.com/key4hep/k4FWCore/blob/338f4ec16d29a8e0d2b2ae98ea781d06fa62d7a8/k4FWCore/src/PodioDataSvc.cpp#L125). 
To make Gaudi Functional use these, we have to make them visible to `[Consumer/Producer/Transformer]`. which internally use `DataHandleMixin`. The latter uses this [template magic](https://gitlab.cern.ch/gaudi/Gaudi/-/blob/master/GaudiFunctional/include/Gaudi/Functional/details.h#L419) to know which Handles to use. To override the default we need to subclass of `Gaudi::Algorithm` that adds something like 
 
```
template<typename T>using InputHandle_t = DataObjectReadHandle<DataWrapper<T>>;
```
and a similar one for the writing part. 

Doing this we will not be able to use the legacy read/write any more though. If we want to preserve that, we have to use the key4hep specific data handle which we wanted to deprecate.

k4FWCore/include/k4FWCore/FunctionalUtils.h
```diff
@@ -0,0 +1,19 @@
...
+
+#include "GaudiAlg/GaudiAlgorithm.h"
+#include "GaudiKernel/DataObjectHandle.h"
+#include "k4FWCore/DataWrapper.h"
+
+// Base class used for the Traits template argument of the
+// Gaudi::Functional algorithms
+struct BaseClass_t {
+  template<typename T>
+  using OutputHandle = DataObjectWriteHandle<DataWrapper<T>>;
```
tmadlener, line 12: Would this still make it possible to produce things that are note a `podio::CollectionBase`? E.g. producers that create (transient) values. Or would we have to do some SFINAE/concept magic here to select the correct Handle here?

test/k4FWCoreTest/src/components/ExampleFunctionalConsumer.cpp
jmcarcell: I was able to do this for the producers. When trying to do the same thing for the consumers (or transformers since they also have inputs) I get the following error message:
```
ExampleFunction...  ERROR Wrong DataObjectType : The type expected for /Event/MCParticles is AnyDataWrapper<edm4hep::MCParticleCollection> and is different from the one of the object in the store which is DataWrapper<podio::CollectionBase>.
```
that is coming from Gaudi performing a check with a `dynamic_cast`: https://gitlab.cern.ch/gaudi/Gaudi/-/blob/master/GaudiKernel/include/GaudiKernel/DataObjectHandle.h#L56. Options:
- Templated registering of objects, not trivial to do
- Using a wrapper around `Gaudi::Functional::Consumer` (and transformer) that retrieves the collection as `podio::CollectionBase` and then casts to the right type, but I like that our algorithms and the ones from Gaudi look the same
- Maybe it's also possible to have a wrapper that we can use for registering and then wrappers of concrete types can inherit from so that the `dynamic_cast` check would pass.

Ideally I would like it to be like it is in the example producers right now, where we inherit from the corresponding `Gaudi::Functional` and we just have to specify the type of collection that we want.

tmadlener: > Ideally I would like it to be like it is in the example producers right now, where we inherit from the corresponding Gaudi::Functional and we just have to specify the type of collection that we want.

I agree with this. In case it is not possible the option I would like the second most after this one is to have a `k4FWCore::Functional` (naming up for discussion) that is effectively a template specialization of the corresponding `Gaudi::Functional` that injects the necessary types for us. I don't know if this is something that is possible, or if would always need a wrapper.

jmcarcell:
The latest commit [`3b09d32` (#129)](https://github.com/key4hep/k4FWCore/pull/129/commits/3b09d3275cb8964ff3aaf634c75281acbb79578b) simplifies the input interfaces allowing not to have to do dynamic casts and not having to use DataWrappers, the price to pay is having to template the registering of objects and having to have explicitly the EDM4hep classes. For classes that are not in EDM4hep they are registered as `podio::CollectionBase` so using `DataWrapper` and dynamic casts will be needed. This is one way of doing this, there are others possibly but this is probably from the user interface the least invasive being it almost identical to a non-k4FWCore Gaudi functional algorithm.

k4FWCore/components/PodioInput.cpp
```diff
@@ -21,21 +21,172 @@
 
 #include "k4FWCore/PodioDataSvc.h"
 
+#include "edm4hep/MCParticleCollection.h"
```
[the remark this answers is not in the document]
jmcarcell: The problem is that we have to register objects with the type we want to get, otherwise there is the check I have mentioned in one of the comments above that will try to dynamic_cast whatever is in the store (`DataWrapper<collectionType>`) to what we are asking for. Pushing a `DataWrapper<podio::CollectionBase>` like before without templating the registration and getting back a `AnyDataWrapper<class>` doesn't work since you can't dynamic_cast one into the other

README.md
```diff
@@ -61,12 +61,29 @@ print(my_opts[0].foo)
...
+``` bash
+mkdir build
+cd build
 cmake ..
 make install
 ```
 
-
+## Implementing algorithms
+k4FWCore uses (or will use) `Gaudi::Functional` for executing algorithms. In
```
hegner, line 72: I'd remove the "or will use". And then: "There are three different types of algorithms, depending on your use case:" And then the enumeration... if you open a ticket on me I can make the following text a bit more explicit - the complicated part is not in the 'operator' part, but in the templating. Which I would add.

README.md
```diff
@@ -61,12 +61,29 @@ print(my_opts[0].foo)
...
+  outputs
+
+In all cases the implementation process is the same, we'll create a new class
+that will implement `operator()`, that is where our algorithm will be. Simple
+examples can be found in the test folder for each one of these three. In
+addition, there are tests that have either multiple inputs and / or multiple
+outputs (like `ExampleFunctionalProducerMultiple`) that can be used as a
+template for the more typical case when working with multiple inputs or outputs.
+
+`GaudiAlg` is deprecated and will be removed in future versions of Gaudi.
```
hegner, line 89: emphasize it visually

k4FWCore/include/k4FWCore/BaseClass.h
```diff
@@ -0,0 +1,19 @@
+#ifndef K4FWCORE_FUNCTIONALUTILS_H
+#define K4FWCORE_FUNCTIONALUTILS_H
+
+#include "GaudiAlg/GaudiAlgorithm.h"
+#include "GaudiKernel/DataObjectHandle.h"
+#include "k4FWCore/DataWrapper.h"
+
+// Base class used for the Traits template argument of the
+// Gaudi::Functional algorithms
+struct BaseClass_t {
```
hegner, line 10: This is a badly chosen name - and not even name spaced. Do you have something more prescriptive what the purpose is?

k4FWCore/src/PodioDataSvc.cpp
```diff
@@ -147,16 +149,12 @@ PodioDataSvc::PodioDataSvc(const std::string& name, ISvcLocator* svc) : DataSvc(
 /// Standard Destructor
 PodioDataSvc::~PodioDataSvc() {}
 
-StatusCode PodioDataSvc::readCollection(const std::string& collName) {
-  const podio::CollectionBase* collection(nullptr);
-  collection = m_eventframe.get(collName);
-  if (collection == nullptr) {
+const std::string_view PodioDataSvc::getCollectionType(const std::string& collName) {
+  auto coll = m_eventframe.get(collName);
```
hegner, line 153: why not const?

test/k4FWCoreTest/src/components/ExampleFunctionalConsumerMultiple.cpp
```diff
@@ -0,0 +1,83 @@
...
+                 }) {}
+
+  // This is the function that will be called to transform the data
+  // Note that the function has to be const, as well as the collections
+  // we get from the input
+  void operator()(
+                  const FloatColl& floatVector,
+                  const ParticleColl& particles, const SimTrackerHitColl& simTrackerHits,
+                  const TrackerHitColl& trackerHits, const  TrackColl& tracks) const override {
+    // if ((floatVector[0] != 125) || (floatVector[1] != 25) || (floatVector[2] != m_event)) {
```
hegner, line 47: I guess you forgot removing

test/k4FWCoreTest/src/components/ExampleFunctionalTransformer.cpp
```diff
@@ -0,0 +1,43 @@
...
+    auto  coll_out = edm4hep::MCParticleCollection();
+    for (const auto& particle : input) {
+      auto new_particle = edm4hep::MutableMCParticle();
+      new_particle.setPDG(particle.getPDG() + 10);
+      new_particle.setGeneratorStatus(particle.getGeneratorStatus() + 10);
+      new_particle.setSimulatorStatus(particle.getSimulatorStatus() + 10);
+      new_particle.setCharge(particle.getCharge() + 10);
+      new_particle.setTime(particle.getTime() + 10);
+      new_particle.setMass(particle.getMass() + 10);
+      // new_particle.
```
hegner, line 36: remove this line

hegner (changes requested):
Thanks. Very nice examples you put. I left a few more comments to address

test/k4FWCoreTest/src/components/ExampleFunctionalTransformer.cpp
jmcarcell: Done

test/k4FWCoreTest/src/components/ExampleFunctionalConsumerMultiple.cpp
jmcarcell: Uncommented

k4FWCore/src/PodioDataSvc.cpp
jmcarcell: Now it's const

k4FWCore/include/k4FWCore/BaseClass.h
jmcarcell: Here the idea is that the Gaudi examples always use a `BaseClass_t` that is defined to be typically `Gaudi::Algorithm` so by having this the changes are minimal. See for example https://gitlab.cern.ch/gaudi/Gaudi/-/blob/master/GaudiExamples/src/FunctionalAlgorithms/MakeAndConsume.cpp#L60 and a few lines below; In k4FWCore we don't need line 60 but we need instead `#include "BaseClass.h"` and the rest is exactly the same as in the Gaudi examples. We could also name it differently and have it namespaced but I thought for consistency it would be simpler this way.

README.md
jmcarcell: I have added a few more comments and a link to a more complete list that I found in the LHCb starterkit. If you want to change it I think you should be able to push to this branch

README.md
jmcarcell: Do you mean add an example here? Or at least the structure with the class definition and template arguments and `operator()`?

test/k4FWCoreTest/src/components/ExampleFunctionalConsumer.cpp
jmcarcell: The `maybeRead` is there to simplify the interface and not have to use `podio::CollectionBase` as type and then do a dynamic cast by pushing directly, for example, an `edm4ep::MCParticleCollection` to the TES. There is maybe an option to create a custom Functional base class we can inherit from in k4FWCore that somehow gets around this so that it is possible to retrieve a `podio::CollectionBase` and do the relevant dynamic casting inside the class and not have the `maybeRead` stuff, but I don't think this is trivial to do

jmcarcell:
As discussed in a private meeting, and for the sake of time for a coming conference and tutorial, this PR will be merged by the end of today if there aren't any complains. pre-commit and the builds with the nightlies both are happy, so it should be ready to go

test/k4FWCoreTest/src/components/ExampleFunctionalConsumer.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include "edm4hep/MCParticleCollection.h"
+
+// Define BaseClass_t
+#include "k4FWCore/BaseClass.h"
+
+#include <stdexcept>
+#include <string>
+
+// Which type of collection we are reading
+using colltype = edm4hep::MCParticleCollection;
```
tmadlener, line 32: Can we remove this typedef? It's effectively used twice (as far as I can see), and it reduces readability IMHO. Additionally, there are other examples with multiple collections that show the usage with a typedef which are much clearer, I think.

If not removing it, at least we should give it a more descriptive name.

test/k4FWCoreTest/src/components/ExampleFunctionalProducer.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "Gaudi/Property.h"
+#include "GaudiAlg/Producer.h"
+#include "k4FWCore/BaseClass.h"
+
+#include "edm4hep/MCParticleCollection.h"
+
+#include <string>
+
+// Which collection we are producing
+using colltype = edm4hep::MCParticleCollection;
```
tmadlener, line 29: In a similar fashion as above, I would remove this typedef.

test/k4FWCoreTest/src/components/ExampleFunctionalProducerMultiple.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include "podio/UserDataCollection.h"
+
+#include <string>
+
+// Which type of collections we are producing
+using Float_t         = podio::UserDataCollection<float>;
+using Particle_t      = edm4hep::MCParticleCollection;
+using SimTrackerHit_t = edm4hep::SimTrackerHitCollection;
+using TrackerHit_t    = edm4hep::TrackerHitCollection;
+using Track_t         = edm4hep::TrackCollection;
```
tmadlener, line 37: Nitpicking here: These typedefs use `_t`, but the example above doesn't. I think at least in the examples we should have consistency.

test/k4FWCoreTest/src/components/ExampleFunctionalTransformerMultiple.cpp
```diff
@@ -0,0 +1,95 @@
...
+// Which type of collection we are reading
+using FloatColl         = podio::UserDataCollection<float>;
+using ParticleColl      = edm4hep::MCParticleCollection;
+using SimTrackerHitColl = edm4hep::SimTrackerHitCollection;
+using TrackerHitColl    = edm4hep::TrackerHitCollection;
+using TrackColl         = edm4hep::TrackCollection;
+
+// As a simple example, we'll write an integer and a collection of MCParticles
+using Counter_t  = podio::UserDataCollection<int>;
+using Particle_t = edm4hep::MCParticleCollection;
```
tmadlener, line 43: Same here as above, unless making the in- and outputs follow different conventions was a conscious choice?

tmadlener (commented):
I had a brief look at the examples and I do have some "style" comments.

jmcarcell:
> I had a brief look at the examples and I do have some "style" comments.

All fixed. They have been changed through many iterations and at some point I became so used to them that I stopped noticing they were around

PR #131, opened 2023-07-28
k4run: allow addition of additional parameters from steering files

Zehvogel:
With this change you can add  a parameter from a steering file by simply doing something like
```
parser.add_argument("-f", "--foo", type=int, help="hello world")
```
this is achieved by making the parser global so that it is also passed to the `exec()` call. This might not be the most beautiful solution, but imo its the simplest and the use of implicitly appearing global variables in the current code is already a thing anyway...

BEGINRELEASENOTES
- k4run: allow addition of additional parameters from steering files

ENDRELEASENOTES


jmcarcell:
But to get access to whatever value would be passed you need to have access to at least one more variable that you would have to make global, otherwise how to do you access the value of foo? Note that we already can pass values corresponding to the parameters, check
``` bash
k4run steering_file.py --help
```

Zehvogel:
True oops, I did not try to use the value yet, I will rework this.

Regarding the other thing: I know about the possibility to pass parameters to the algorithms, but I want to have parameters to set which parts of a steering file should run

Zehvogel:
This was quite tricky but the solution was simple.

With these changes it is now possible to add and parse arguments inside of steering files while keeping the help intact. Moreover, it also fixes the bug of 
```
k4run --help file.py
```
not showing the Gaudi properties while
```
k4run file.py --help
```
did.

jmcarcell:
Can you document that it is possible to do it and how to do it?

Zehvogel:
Sure I can document it, can you point me to any `k4run` or steering file documentation where I should add it to?

Zehvogel:
I added some documentation to the README

Zehvogel:
Superseded by: https://github.com/key4hep/k4FWCore/pull/134

Issue #125, opened 2023-07-20
Running over more events than present or all events (using `-1`) should exit with 0


tmadlener:
Currently anything that makes Gaudi use a `IEventProcessor::stopRun` makes the whole process exit with exit code 4 (see [here](https://gitlab.cern.ch/gaudi/Gaudi/-/blob/4784576d0dd1249c456f1057cbd472e1637be20f/GaudiKernel/include/GaudiKernel/AppReturnCode.h#L35)).

This requires us to do things like this in tests:
https://github.com/key4hep/k4FWCore/blob/33ba58758736793d2a4f1dcfe0aa5fb5dd119d1c/test/k4FWCoreTest/CMakeLists.txt#L60-L62

It also makes it impossible to defensively script (using `-e`). This should be fixed somehow, although I do not know how.

PR #132, opened 2023-08-02
k4run: return exit code 0 if `-n -1` was used and Gaudi returned exit code 4

Zehvogel:
Fixes https://github.com/key4hep/k4FWCore/issues/125

BEGINRELEASENOTES
- k4run: return exit code 0 if `-n -1` was used and Gaudi returned exit code 4 (user requested stop)
- Summary of the changes:

Exit code when using `EvtMax != -1`
<table> 
<tr> 
 <td> Running over...
 <td> Previously
 <td> Now
<tr>
 <td> less events than available
 <td> 4
 <td> 0
<tr>
 <td> same events as available
 <td> 4
 <td> 0
<tr>
 <td> more events than available
 <td> 4
 <td> 4
</table>

And when using `EvtMax = -1` previously it was 4, now it's 0.

ENDRELEASENOTES


k4FWCore/scripts/k4run
```diff
@@ -171,4 +171,7 @@ if __name__ == "__main__":
     if not opts.dry_run:
         # Do the real processing
         retcode = c.run(opts.gdb, opts.ncpus)
+        # User requested stop returns non-zero exit code see: https://github.com/key4hep/k4FWCore/issues/125
+        if ApplicationMgr().EvtMax == -1 and retcode == 0x04:
+           retcode = 0x00
```
jmcarcell, line 176: Ugh can we use at least `retcode = 0` instead of `0x00`, the other one I care less because it's coming from Gaudi but it looks so ugly

jmcarcell:
`PodioDataSvc` always stops the run (this is the scheduled stop) when there aren't any more events to read. So when `EvtMax == -1` it will certainly happen but it's also possible when `EvtMax` is greater than the number of events, so to cover both the check on `EvtMax` would have to be removed. It depends if we consider a failure that case or not. One option would be to return 0 but add a warning that we are trying to read more events than there are available. At the moment, you get an INFO message saying that the end of the file was reached, whether EvtMax was -1 or not:
```
EventDataSvc         INFO Reached end of file with event 100
```

k4FWCore/scripts/k4run
Zehvogel: 0x00 is also from Gaudi ;) but sure, I can change them. :)

Zehvogel:
Imo telling Gaudi to read more events than you have in your file should not return 0

andresailer:
I agree: no gratuitous hex, and return error when less events than requested were processed.

jmcarcell:
It's a bit trickier than that because the case when you read the exact number of events in the file also gives you a code 4 and you can't distinguish from `k4run`. So the options are
- This PR: check for -1, then both reading the exact number of events and reading more events would not return 0
- No check, return always 0 when stopping reading and give an error if you try to read more than there are available, maybe an option to make this a warning if you don't care about that from the user side
- 
I'm not sure if stopping the run when you try to read beyond the maximum event (instead of now that if a file has N events and you read the N, that triggers the stop call there, but I think that means don't read the next event) would work, but that would make this PR work

jmcarcell:
Sounds easy but simply modifying the current check to trigger when reading one beyond the last one will lead to problems as the algorithms continue in the current event. And if it's one beyond the current one then there is no event and everything we have under `execute` expects to run on an actual good event

Zehvogel:
Yes looks like it will not be that easy to fix. I have moved one of the tests to cover this case now and added another one testing for -1 again. 

I will convert back to draft for now until I found a solution

Zehvogel:
I have changed this so that the scheduled stop is only issued with `--num-events=-1`.
When requesting to run with more events than available Gaudi crashes (rightfully so I think).

jmcarcell:
I prefer the current behavior where it stops at the last one instead of crashing, there could be lots of places where this happens and works fine and this would break that.

Zehvogel:
The current behaviour masks if you are doing something wrong

tmadlener:
Just to add one more point for discussion here: There might be a legitimate reason for running with more events than available from a file. If you run things on a batch system and you have, e.g. an input file with 102k and you split that into jobs of 10k events each, the last one will only have 2k events, but adjusting for that in the surrounding batch configuration might be rather complicated (or at least introduce overhead there).

Zehvogel:
I can of course remove all the `PodioDataSvc` changes from this and just make it always return 0 when a user requested stop happened if that is wanted. I guess it is somewhat justified as we don't tell Gaudi how many events it should process but only what the _maximum_ number of events to process is :shrug:

jmcarcell:
> I can of course remove all the `PodioDataSvc` changes from this

Yes please. Independently of what we do this makes it crash in some situations while currently it doesn't so that isn't an improvement. Re always 0 or not I don't think I have a strong opinion. @andresailer prefers error code when processing less than requested from one of the first comments

Zehvogel:
Doing this I remember I notice again why I did it the way I did. With just the change in k4run I would also still return 4 for processing exactly the right amount of events. I.e. the only thing we can easily do is always return 0 instead of 4 even when trying to process more events than available. 

Otherwise I would need a way to ask Gaudi afterwards how many events it processed, [unfortunately Gaudi doesn't seem to keep track of that](https://gitlab.cern.ch/atlas/Gaudi/-/blob/master/GaudiCoreSvc/src/ApplicationMgr/EventLoopMgr.cpp#L287)...

andresailer:
I guess I will grudgingly survive, if things are successful even when reading less events than requested?

Zehvogel:
Now it does the following:
If run with `-n m` and enough events exist, no bounds checking is done and Gaudi returns 0 on its own.
In all other cases bounds checking is done and a stop is requested when running out of events.
If `-n -1` we change the return code from 4 to 0.
Else, i.e. when requesting more events than available we return what Gaudi returns i.e. return code 4

k4FWCore/src/PodioDataSvc.cpp
```diff
@@ -62,6 +63,22 @@ StatusCode PodioDataSvc::initialize() {
...
 
+  IProperty* property;
+  auto       sc = service("ApplicationMgr", property);
+  if (sc == StatusCode::FAILURE) {
+    error() << "Could not get ApplicationMgr properties" << std::endl;
+  }
+  Gaudi::Property<int> evtMax;
+  evtMax.assign(property->getProperty("EvtMax"));
+  m_requestedEventMax = evtMax;
+  m_requestedEventMax -= m_1stEvtEntry;
```
tmadlener, line 74: Is going through a dedictated `Gaudi::Property` necessary? Could we not simply do `property->getProperty("EvtMax")` and assign that directly to `m_requestedEventMax`? Is this mainly to get the correct (implicit) conversion to `int`? Can we trigger an explicit conversion from the original property and skip the intermediate Property?

tmadlener (commented):
I think the updated behavior makes sense and at least is what I would expect to happen intuitively. I have one curiosity below, but that is really no blocker at all.

k4FWCore/src/PodioDataSvc.cpp
Zehvogel: I could do this instead if you prefer
```
m_requestedEventMax = dynamic_cast<const Gaudi::Property<int>&>(property->getProperty("EvtMax"));
```

tmadlener: Eh, not entirely better ;) I think keeping the current version is fine.

jmcarcell: Another alternative is 
``` cpp
m_requestedEventMax = std::stoi(property->getProperty("EvtMax").toString());
```

k4FWCore/src/PodioDataSvc.cpp
```diff
@@ -62,6 +63,22 @@ StatusCode PodioDataSvc::initialize() {
...
   }
 
+  IProperty* property;
+  auto       sc = service("ApplicationMgr", property);
+  if (sc == StatusCode::FAILURE) {
+    error() << "Could not get ApplicationMgr properties" << std::endl;
+  }
+  Gaudi::Property<int> evtMax;
+  evtMax.assign(property->getProperty("EvtMax"));
+  m_requestedEventMax = evtMax;
```
jmcarcell, line 73:
```suggestion
  m_requestedEventMax = std::stoi(property->getProperty("EvtMax").toString());
```
I liked the one liner

k4FWCore/src/PodioDataSvc.cpp
```diff
@@ -109,10 +126,16 @@ StatusCode PodioDataSvc::i_setRoot(std::string root_path, DataObject* pRootObj)
 }
 
 void PodioDataSvc::endOfRead() {
+  m_eventNum++;
+
+  if (!m_bounds_check_needed) {
+    return;
+  }
```
jmcarcell, line 133:
```suggestion

  if (!m_bounds_check_needed) {
    return;
  }
  
    m_eventNum++;
```
if it's going to return we don't need to increment anything

k4FWCore/src/PodioDataSvc.cpp
```diff
@@ -109,10 +126,16 @@ StatusCode PodioDataSvc::i_setRoot(std::string root_path, DataObject* pRootObj)
...
+
   StatusCode sc;
-  if (m_eventMax != -1) {
-    if (m_eventNum++ >= m_eventMax - 1) {  // we start counting at 0 thus the -1.
-      info() << "Reached end of file with event " << m_eventMax << endmsg;
+  if (m_availableEventMax != -1) {
+    if (m_eventNum >= m_availableEventMax - 1) {  // we start counting at 0 thus the -1.
```
jmcarcell, line 137:
```suggestion
    if (m_availableEventMax != -1 && m_eventNum + 1>= m_availableEventMax) {  // this will process the current event thus the +1.
```
(plus removing a } that I can't suggest)
I think this comment is better since it's not very intuitive because what happens is that the current event gets processed so you have to check if the next one will be beyond bounds

k4FWCore/src/PodioDataSvc.cpp
```diff
@@ -109,10 +126,16 @@ StatusCode PodioDataSvc::i_setRoot(std::string root_path, DataObject* pRootObj)
 }
 
 void PodioDataSvc::endOfRead() {
+  m_eventNum++;
+
+  if (!m_bounds_check_needed) {
+    return;
+  }
+
   StatusCode sc;
```
jmcarcell, line 135: This can be inside the loop, there is also a comment about this being unused that I don't think is relevant anymore

jmcarcell:
I think I made all my comments, other than that it looks good. We can change the tests that don't need to pass using a regular expression in this or another PR.
This table should summarize the status and changes
<table>
<tr>
 <td>	
 <td> Previously
 <td> Now
<tr>
 <td> Less events than available
 <td> 0
 <td> 0
<tr>
 <td> Same events as available
 <td> 4
 <td> 0
<tr>
 <td> More events than available
 <td> 4
 <td> 4
</table>

tmadlener:
Thanks for the summary table. I would put that into the release notes, so that people don't have to scroll all the way down (or even have to go to the PR for the details).

k4FWCore/src/PodioDataSvc.cpp
Zehvogel: that looks to me like the least clean way to do this... :D

k4FWCore/src/PodioDataSvc.cpp
Zehvogel: We always need to increment the event number (at least if we aren't in the last event yet). This change would cause us to read the same event again and again.

k4FWCore/src/PodioDataSvc.cpp
Zehvogel: I could put this inside of the if yes, I just did not want to make more changes than necessary. 
The comment is still valid I think. The variable is only there to hold the `StatusCode`(s) returned by
```
      sc = service("ApplicationMgr", eventProcessor);
      sc = eventProcessor->stopRun()
```
I guess we should check them and act accordingly, but I gladly leave that change to someone else.

k4FWCore/src/PodioDataSvc.cpp
jmcarcell: Ah indeed, I thought https://github.com/AIDASoft/podio/blob/master/include/podio/ROOTFrameReader.h#L83 was being used, I think that would be better than having to worry about counting since it's always going to be the next one

k4FWCore/src/PodioDataSvc.cpp
jmcarcell: Well it is correct, the current event that `m_eventNum` is pointing to will be processed unconditionally so that's why we have to check for the next event after the one that `m_eventNum` (maybe this is not explicit enough) is pointing to independently of the value that it has right now. That's why I feel the +1 is easier to understand than the -1.

Zehvogel: > Well it is correct, the current event that `m_eventNum` is pointing to will be processed unconditionally

No?

k4FWCore/src/PodioDataSvc.cpp
Zehvogel: Even then we would still need to do the current counting as we need to issue the stop to gaudi before we actually hit the non-existing event as far as I understand

k4FWCore/src/PodioDataSvc.cpp
```diff
@@ -109,14 +126,20 @@ StatusCode PodioDataSvc::i_setRoot(std::string root_path, DataObject* pRootObj)
...
-  if (m_eventMax != -1) {
-    if (m_eventNum++ >= m_eventMax - 1) {  // we start counting at 0 thus the -1.
-      info() << "Reached end of file with event " << m_eventMax << endmsg;
-      IEventProcessor* eventProcessor;
-      sc = service("ApplicationMgr", eventProcessor);
-      sc = eventProcessor->stopRun();
-    }
+  // m_eventNum points to the next event already so we check if its available
+  // (m_eventNum starts counting from 0, so <)
+  if (!(m_eventNum < m_availableEventMax)) {
```
Zehvogel, line 138: @jmcarcell how do you like this?

jmcarcell: I don't think they are equivalents, if `m_availableEventMax == -1` then the new version can trigger while the old one wouldn't. The check is fine, I don't think we need to change it, it's just the commenting and naming

Zehvogel: true :/ I will roll this back

k4FWCore/src/PodioDataSvc.cpp
```diff
@@ -109,14 +126,19 @@ StatusCode PodioDataSvc::i_setRoot(std::string root_path, DataObject* pRootObj)
...
-  if (m_eventMax != -1) {
-    if (m_eventNum++ >= m_eventMax - 1) {  // we start counting at 0 thus the -1.
-      info() << "Reached end of file with event " << m_eventMax << endmsg;
-      IEventProcessor* eventProcessor;
-      sc = service("ApplicationMgr", eventProcessor);
-      sc = eventProcessor->stopRun();
-    }
+  // m_eventNum already points to the next event here so check if it is available
+  if (m_numAvailableEvents != -1 && m_eventNum >= m_numAvailableEvents) {
+    info() << "Reached end of file with event " << m_eventNum << "/" << m_numAvailableEvents << endmsg;
```
Zehvogel, line 138: Turns out the -1 was just wrong, I noticed it when I extended this info message.
With the -1 we stop at e.g. `m_eventNum == 99` when 100 events are available. This would be correct because `m_eventNum` starts at 0, if `m_eventNum` would not already point to the next event.

Zehvogel:
let me know when you are happy to merge, then I will transfer the same changes to the legacy service and add tests for that one as well

k4FWCore/src/PodioDataSvc.cpp
Zehvogel: I am also not so sure if we can even hit the "m_numAvailableEvents != -1" check, for that we would need a file not containing any events and I think than we don't even reach the endOfRead check

k4FWCore/src/PodioDataSvc.cpp
```diff
@@ -43,10 +44,10 @@ StatusCode PodioDataSvc::initialize() {
     if (m_filenames[0] != "") {
       m_reading_from_file = true;
       m_reader.openFiles(m_filenames);
-      m_eventMax = m_reader.getEntries("events");
+      m_numAvailableEvents = m_reader.getEntries("events");
 
       if (m_1stEvtEntry != 0) {
-        m_eventMax -= m_1stEvtEntry;
+        m_numAvailableEvents -= m_1stEvtEntry;
```
tmadlener, line 50: This `if` condition has no effect, since `-= 0` will also leave `m_numAvailableEvents` unchanged.

k4FWCore/src/PodioDataSvc.cpp
tmadlener: Looking through the code, is `m_numAvailableEvents` ever still `-1` here? At least from a quick glance I couldn't see a path where it is not assigned a non-default value(?).

tmadlener (commented):
I just have some nitpicky comments that mainly caught my eye because of the new variable (name).

k4FWCore/src/PodioDataSvc.cpp
tmadlener: Even if there is a file without events, `reader.getEntries` will return 0. Things might get wonky if there is a file with `numeric_limits<unsigned>::max()` events, as that could result in `-1` again ;)

jmcarcell: > With the -1 we stop at e.g. `m_eventNum == 99` when 100 events are available.

Good this is what I noticed the last day but haven't had time to reproduce it since. I'll have a final look today to see if it behaves correctly always

k4FWCore/src/PodioDataSvc.cpp
```diff
@@ -109,14 +123,19 @@ StatusCode PodioDataSvc::i_setRoot(std::string root_path, DataObject* pRootObj)
...
-  if (m_eventMax != -1) {
-    if (m_eventNum++ >= m_eventMax - 1) {  // we start counting at 0 thus the -1.
-      info() << "Reached end of file with event " << m_eventMax << endmsg;
-      IEventProcessor* eventProcessor;
-      sc = service("ApplicationMgr", eventProcessor);
-      sc = eventProcessor->stopRun();
-    }
+  // m_eventNum already points to the next event here so check if it is available
+  if (m_eventNum >= m_numAvailableEvents) {
+    info() << "Reached end of file with event " << m_eventNum << "/" << m_numAvailableEvents << endmsg;
```
jmcarcell, line 135: I hate to keep adding on this but this comment won't run when your request to read exactly the number of events that a file has, only with `EvtMax = -1` or when you request to read more events than available. Therefore it's a bit of redundant information because when it will appear it will always print "Reached ... N/N" where N is the number of events

Zehvogel: do you have a change suggestion?

jmcarcell: Not really, I only think it would be useful when you request more events than available and it tells you that it has stopped. Other than that sometimes it may appear (with `EvtMax = -1`) or not (when reading N events with N available events) when reaching the end of the file so it's a bit inconsistent. It would be possible by moving the info message above the return in `endOfRead()` so that it also runs in that last case when reading N with N available

Zehvogel: I changed it to a more meaningful message:
"Reached end of file with event N (M events requested)"

k4FWCore/src/PodioDataSvc.cpp
```diff
@@ -109,14 +123,20 @@ StatusCode PodioDataSvc::i_setRoot(std::string root_path, DataObject* pRootObj)
...
-  if (m_eventMax != -1) {
-    if (m_eventNum++ >= m_eventMax - 1) {  // we start counting at 0 thus the -1.
-      info() << "Reached end of file with event " << m_eventMax << endmsg;
-      IEventProcessor* eventProcessor;
-      sc = service("ApplicationMgr", eventProcessor);
-      sc = eventProcessor->stopRun();
-    }
+  // m_eventNum already points to the next event here so check if it is available
+  if (m_eventNum >= m_numAvailableEvents) {
+    info() << "Reached end of file with event " << m_eventNum << "(" << m_requestedEventMax << " events requested)"
```
jmcarcell, line 135:
```suggestion
    info() << "Reached end of file with event " << m_eventNum << " (" << m_requestedEventMax << " events requested)"
```

PR #135, opened 2023-08-11
Allow help without config file

kjvbrt:
BEGINRELEASENOTES

- Allow help without config file

ENDRELEASENOTES

@Zehvogel can this ugly if be avoided?

### DIFF
--- a/k4FWCore/scripts/k4run
+++ b/k4FWCore/scripts/k4run
@@ -122,7 +122,8 @@ if __name__ == "__main__":
                   print("  %s (from %s)" % (item, cfgDb[item]["lib"]))
         sys.exit()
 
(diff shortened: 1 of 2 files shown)


Zehvogel:
Good job spotting this!
Sorry I only added the if last minute to fix the running k4run without arguments test and missed that this breaks something else...
Can you add a test to this PR that checks if `k4run --help` works so that we avoid such an accident in the future?

I can try to come up with something better than this if..

Zehvogel:
instead of adding to the if it might be more elegant to simply move the whole if down to e.g. line 137 and change `opts[0]` to `opts`

Zehvogel (approved):
LGTM, thx!

tmadlener (approved):

PR #136, opened 2023-08-11
Ignoring root files generated by tests

kjvbrt:
BEGINRELEASENOTES

- Ignoring root files generated by tests

ENDRELEASENOTES

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -241,4 +241,5 @@ edm4hep/src/
 ## k4MarlinWrapper
 test/inputFiles/*.slcio
 test/gaudi_opts/testConverterConstants.py
-
+# Ignore root files generated in tests
+test/**/*root


Zehvogel:
Good idea, I think @jmcarcell  also planned to deploy a centrally managed `.gitignore`to all repos from somewhere, we should take care to also get this into that.

.gitignore
```diff
@@ -241,4 +241,5 @@ edm4hep/src/
 ## k4MarlinWrapper
 test/inputFiles/*.slcio
 test/gaudi_opts/testConverterConstants.py
-
+# Ignore root files generated in tests
+test/**/*root
```
tmadlener, line 245: Do we have any root files that need to be tracked? Otherwise we could also simply do
```suggestion
*.root
```

kjvbrt: I would limit the ignore to just what is necessary, similarly to slcio files few lines above.

jmcarcell:
Yeah I forgot about those root files, there may be more in other repos. I just synced the `.gitignore` from https://github.com/key4hep/key4hep-dev-utils/blob/main/defaults/.gitignore in all the repos. If that's fine we can close this PR

PR #143, opened 2023-09-01
Fix clang warnings by adding final

jmcarcell:
BEGINRELEASENOTES
- Fix clang warning inconsistent-missing-override by adding final

ENDRELEASENOTES

Supersedes https://github.com/key4hep/k4FWCore/pull/96 (fixes a few more warnings)


jmcarcell:
I know we have more places with a similar structure but I only get warnings for these. It seems that those that inherit from `GaudiAlgorithm` don't get the warnings for some reason, while the `DataSvcs` do.

Issue #105, opened 2023-05-22
Remove the reader from the `PodioDataSvc` and make the `PodioInput` populate the internal Frame


tmadlener:
Currently, the `PodioDataSvc` holds a `ROOTFrameReader` which is used for configuring the input file. On the other hand the collections that should be read are configured via the `PodioInput` algorithm. All of the input configuration should live in the `PodioInput`.

Other points to address:
- Currently all collections have to be stated in the options file. This should become optional to restrict the collections with a default of reading all collections if not stated otherwise.

Food for thought: Could we simply take the default `DataHandle` from Gaudi and populate it in `PodioInput` (from the Frame we read there) and take things out again in `PodioOutput` (put it into a Frame and write things out). Could potentially solve the issues we observe with [k4Gen](https://github.com/key4hep/k4FWCore/pull/100#issuecomment-1556734077).

jmcarcell:
This is probably not going to happen since they are going to be deprecated at some point, see https://github.com/key4hep/k4FWCore/issues/292.

PR #145, opened 2023-09-20
Add support for reading any number of collections with Gaudi::Functional

jmcarcell:
Continuation of https://github.com/key4hep/k4FWCore/pull/129

Adds support to read any number of collections for cases when how many collections are going to be read is specified at runtime in the steering files. For example, let's say we want to process all the calorimeter hits and we have several collections with calorimeter hits so with this PR all of them can be read at once and processed in the same algorithm (I don't think this is supported right now in the pre-functional k4FWCore).

The collections are specified as a list of space separated collections to `PodioInput`, like `Coll1 Coll2 Coll3`, and the algorithm receives a `std::map<std::string, CollType*>` mapping each collection name to the type in case something different is done based on the collection name (the type of all of them should be the same). The price to pay as in #129 for the interface without wrappers and `podio::CollectionBase` is to have a templated registering of every map-type for now.

Whether https://github.com/key4hep/k4FWCore/pull/129 is merged in its current state or not, we should support this way of reading.

This PR doesn't need to be merged (another PR could be made from this branch to main) but it is done this way so the diff does not show all the functional changes.

### DIFF
--- a/k4FWCore/include/k4FWCore/PodioDataSvc.h
+++ b/k4FWCore/include/k4FWCore/PodioDataSvc.h
@@ -68,14 +68,48 @@ public:
 
   const std::string_view getCollectionType(const std::string& collName);
 
-  template <typename T> StatusCode readCollection(const std::string& collName) {
-    const T* collection(nullptr);
-    collection = static_cast<const T*>(m_eventframe.get(collName));
-    if (collection == nullptr) {
+  template<typename T>
+  struct is_map
+  {
+      static constexpr bool value = false;
+  };
+
+
+  template <typename... Args>
+  struct is_map<std::map<Args...>>
+  {
+      static constexpr bool value = true;
+  };
+
+  template <typename T>
+  std::enable_if_t<!is_map<T>::value,StatusCode>
+  readCollection(const std::string& collName) {
+    const T* collection = static_cast<const T*>(m_eventframe.get(collName));
+    if (!collection) {
       error() << "Collection " << collName << " does not exist." << endmsg;
     }
     auto wrapper = new DataWrapper<T>;
     wrapper->setData(collection);
+    m_podio_datawrappers.push_back(wrapper);
+    return DataSvc::registerObject("/Event", "/" + collName, wrapper);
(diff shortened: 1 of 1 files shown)


tmadlener:
If I understand correctly this still requires the user to state all the collections that need to be read, right? I have mentioned this in https://github.com/key4hep/k4FWCore/issues/105 but I would really like to remove that necessity, so that `PodioInput` by default reads all the collections and if users want to **restrict** this to just a subset they can use a list to do so. Do you think something like this is possible?

tmadlener:
After thinking about this a bit more, I am not sure this is the best way to go about this. Just for my understanding the main use case for this, IIUC, is to be able have algorithms that operate on more than one input collection, right? Hence, the `map` that is introduced is effectively never used as an associative container for lookup, but rather just to hold multiple (name, collection) pairs to loop over?

Would it then not be easier to have a "generic" (to whatever degree possible) algorithm that simply "merges" collections? That would also allow us to easily have a list of strings as a parameter rather than a space separated string that we have to parse into a list. This would effectively take all desired collections (of the same type) and simply create a subset collection as output. This subset collection can then be used in algorithms afterwards. It is in principle a fully "virtual collection" as it should also be easy to simply drop it from the output without breaking any of the relations / associations.

Then another unrelated comment to the above. This registering of the functions with the correct types makes things work with EDM4hep, right? But what happens if someone has extended EDM4hep and now wants to use their datatypes with Gaudi::Functional? Can that be easily solved?

andresailer:
> Would it then not be easier to have a "generic" (to whatever degree possible) algorithm that simply "merges" collections

For the ConformalTracking we want to give a list of collections, and treat different collections differently.
I.e.: Start with VertexBarrel, Extend to VertexEndcap. Which is not something we can do if all hits are in one collection. At least not by name of the collection, but instead sub-detector system ID probably, or something like that.

tmadlener:
Is this still relevant after #129 has been merged?

jmcarcell:
This is probably not going to be the way this is implemented...

PR #146, opened 2023-09-21
Adapt to new optional return values from collID table

tmadlener:
BEGINRELEASENOTES
- Adapt the usage of the CollectionIDTable in the Legacy Podio In/Output algorithms to the new interface of the CollectionIDTable, introduced in https://github.com/AIDASoft/podio/pull/486

ENDRELEASENOTES


jmcarcell:
Merging since https://github.com/AIDASoft/podio/pull/486 broke the nightlies without this

tmadlener:
Ah sorry. We were too quick on the Friday then. I didn't think about the break here.

PR #155, opened 2023-10-24
Clean up and rewrite the tests in cmake

jmcarcell:
BEGINRELEASENOTES
- Clean up and rewrite the tests in cmake. Compile all the .cpp files instead of specifying them manually. Define a function to set the environment so that adding a test is a one liner in most cases.

ENDRELEASENOTES

I would like one test = one line, currently we have way too much repetition.


jmcarcell:
-140 lines of cmake, formatting will be left for another time; to be done centrally like with `.clang-format`

PR #159, opened 2023-11-09
Make PodioOutput exit gracefully instead of crashing

### LINKED ISSUE #68: PodioOutput will fail when writing collections only in specific events
- See PR: https://github.com/key4hep/k4FWCore/pull/67
- If a collection is written only in one event, but not another, it will break

- It breaks on filling the `TTree* m_datatree`

tmadlener:
BEGINRELEASENOTES
- Catch the exception that was introduced in AIDASoft/podio#513 and report failure in PodioOutput. 

ENDRELEASENOTES

Fixes #68 (by making sure inconsistent events are impossible)
Fixes https://github.com/key4hep/k4MarlinWrapper/issues/94 (by avoiding the segfault)

- [ ] Needs AIDASoft/podio#513


tmadlener:
Unless there is anything that requires changes for AIDASoft/podio#513 from this side, I would merge that (aiming for early next week).

andresailer (approved):

jmcarcell (approved):

k4FWCore/components/PodioOutput.cpp
```diff
@@ -57,7 +57,25 @@ StatusCode PodioOutput::execute() {
...
+      // In this error message we are only interested in the ones that are
+      // missing, since only a missing collection can trigger the exception
+      // here. Additional collections that are present in the Frame are not
+      // necessarily an issue here, because we might just be configured to not
+      // write all of them
+      const auto& [missing, _] = m_framewriter->checkConsistency(frame.getAvailableCollections(), "events");
+      error() << "Could not write event, because the following collections are not present: ";
+      std::string sep = "";
+      for (const auto& n : missing) {
+        error() << sep << n;
```
jmcarcell, line 72:
```suggestion
      for (const auto& name : missing) {
        error() << sep << name;
```

jmcarcell: For readability

jmcarcell: For readability

PR #160, opened 2023-11-14
PodioInput: always read EventHeader

Zehvogel:
BEGINRELEASENOTES
- PodioInput: always read the EventHeader

ENDRELEASENOTES

see discussion in https://github.com/key4hep/k4MarlinWrapper/pull/162

after https://github.com/key4hep/EDM4hep/pull/239 I will upgrade this to use the new constant

PR #161, opened 2023-11-16
Make "always read EventHeader" collection more robust

tmadlener:
BEGINRELEASENOTES
- Make sure that the EventHeader is always read when it is available, but also to not read it if it is not available.
- Add a test case that checks that the `EventHeaderFiller` (introduced in #133) are correct when reading back.
- Make sure that the `PodioDataSvc` only reads and registers collections to the EDS once even in case there are duplicates.

ENDRELEASENOTES

#160 introduced a bug that went undetected in tests, because all the tests apparently override the list of collections that is set in the `PodioInput` constructor. Using an empty (default) list and an input file that doesn't contain the `EventHeader` collection leads to Gaudi exiting early because it cannot retrieve the collection from the event store.

This PR adds a simple consumer that makes sure that things run, even if the input file does not have an EventHeader collection.

Overall there is a discussion to be head whether making the `EventHeader` a datatype like all the rest is really the best design, as it is technically rather part of the "metadata" that describe an Event rather than content of it.

- [x] Actually implement the fix
- [x] Release Notes


k4FWCore/components/PodioInput.cpp
```diff
@@ -184,14 +184,15 @@ PodioInput::PodioInput(const std::string& name, ISvcLocator* svcLoc) : Consumer(
...
-
-  auto key = edm4hep::EventHeaderName;
-  if (std::find(m_collectionNames.begin(), m_collectionNames.end(), key) == m_collectionNames.end()) {
-    m_collectionNames.value().push_back(key);
-  }
 }
 
 void PodioInput::operator()() const {
+  if (m_podioDataSvc->hasCollection(edm4hep::EventHeaderName)) {
+    m_readers[edm4hep::EventHeaderCollection::typeName](edm4hep::EventHeaderName);
```
Zehvogel, line 191: I thought about doing it this way initially, what happens if the EventHeader collection is also in the list of collections to read? Will it be read twice?

tmadlener: It would query the underlying Frame again for a second time. However, the second time around that effectively is a map lookup and return. Technically, `hasCollection` does all the "heavy lifting" of prompting the Frame to actually read the data and populate it's internal map. Most importantly you will always get the same collection.

I quickly checked what happens if I explicitly make the test read the EventHeader collection and nothing seems to have broken, but it looks like we would be putting two DataHandles pointing to the same collection into the EDS at the moment. I have to check why that seems to work without breaking.

tmadlener: We indeed do get two `DataHandles` (and corresponding objects in the EDS) if `EventHeader` is explicitly passed in. However, since the collection is owned by the Frame and the DataSvc defers the cleanup to that, there is at least no double free.

I will add a simple filter to clean out duplicate collection names as well as `"EventHeader"`. Additionally, I will adapt the `PodioDataSvc::readCollection` to check for existence in the EDS before populating it twice.

tmadlener: Didn't do the filtering in the end, because it is effectively unnecessary with the guard against duplicate entries in the EDS.

tmadlener:
Unless there are any complaints until this evening I would merge this and then tag k4FWCore

Zehvogel (approved):

test/k4FWCoreTest/CMakeLists.txt
```diff
@@ -95,6 +95,7 @@ add_test_with_env(TestDataHandleUniquePtr options/TestDataHandleUniquePtr.py)
 add_test_with_env(TestUniqueIDGenSvc options/TestUniqueIDGenSvc.py)
 add_test_with_env(CreateLegacyExampleEventData options/createLegacyExampleEventData.py)
 add_test_with_env(TestEventHeaderFiller options/createEventHeader.py)
+add_test_with_env(EventHeaderCheck options/runEventHeaderCheck.py PROPERTIES DEPENDS TestEventHeaderFiller)
```
jmcarcell, line 98: Look at how easy it is to add one test now...

test/k4FWCoreTest/src/components/ExampleEventHeaderConsumer.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+#include <Gaudi/Property.h>
+#include <GaudiAlg/Consumer.h>
+#include <GaudiKernel/ISvcLocator.h>
+
+#include <fmt/core.h>
+#include <fmt/format.h>
+
+#include <atomic>
+#include <optional>
```
jmcarcell, line 33: Not needed?

tmadlener: Right, good catch. I was experimenting with optional inputs in Gaudi::Functional and forgot to remove this.

tmadlener, line 33:
```suggestion
```

k4FWCore/components/PodioInput.cpp
```diff
@@ -184,14 +184,15 @@ PodioInput::PodioInput(const std::string& name, ISvcLocator* svcLoc) : Consumer(
...
   fillReaders();
-
-  auto key = edm4hep::EventHeaderName;
-  if (std::find(m_collectionNames.begin(), m_collectionNames.end(), key) == m_collectionNames.end()) {
-    m_collectionNames.value().push_back(key);
-  }
 }
 
 void PodioInput::operator()() const {
+  if (m_podioDataSvc->hasCollection(edm4hep::EventHeaderName)) {
```
jmcarcell, line 190: Since it only happens one the `hasCollection` function can be removed, I think this should work:

```suggestion
  if (m_podioDataSvc->getEventFrame.get(edm4hep::EventHeaderName)) {
```

jmcarcell:
I don't have any more comments; I think many parts need to be reworked anyway to be able to use the `PodioHiveWhiteBoard` so that is something we can change after the release

k4FWCore/components/PodioInput.cpp
tmadlener: Done.

PR #162, opened 2023-11-17
PodioInput: Read all collections by default, allow to limit them with the `collections` property

tmadlener:
BEGINRELEASENOTES
- Make the PodioInput read all collections by default, instead of requiring a full list of collections in the python config.
  - Keep the `collections` property to allow limiting the collections that are read to a subset of collections.
- Avoid crashing with a seg-fault in case a non-existent collection is requested and instead emit a warning and continue. 

ENDRELEASENOTES

- [x] Includes #161 
- [x] Tests

Partially addresses #105

### DIFF
--- a/k4FWCore/components/PodioInput.cpp
+++ b/k4FWCore/components/PodioInput.cpp
@@ -186,6 +186,17 @@ PodioInput::PodioInput(const std::string& name, ISvcLocator* svcLoc) : Consumer(
   fillReaders();
 }
 
+StatusCode PodioInput::initialize() {
+  // If someone uses the collections property from the command line and passes
+  // an empty string we assume they want all collections (as a simple way to
+  // override whatever is in the options file)
+  if (m_collectionNames.size() == 1 && m_collectionNames[0].empty()) {
+    m_collectionNames.clear();
+  }
+
+  return StatusCode::SUCCESS;
+}
+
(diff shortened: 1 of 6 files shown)


k4FWCore/components/PodioInput.cpp
```diff
@@ -184,16 +184,29 @@ PodioInput::PodioInput(const std::string& name, ISvcLocator* svcLoc) : Consumer(
...
+    info() << "No EventHeader collection found in the event. Not reading it" << endmsg;
+  }
+
+  const auto& collsToRead = [&]() {
+    if (m_collectionNames.empty()) {
+      return m_podioDataSvc->getEventFrame().getAvailableCollections();
+    } else {
+      return m_collectionNames.value();
+    }
+  }();
```
Zehvogel, line 202: my c++ knowledge is not as advanced but why do we need this as a lambda? :woozy_face:

tmadlener: The pattern is called an *immediately invoked function expression (IIFE)* and it is mainly used for "complicated" initialization logic, where it can help to
- reduce the scope of temporary variables that are necessary for the initialization
- **Allows to declare the initialized object as `const`**, instead of having it mutable and dealing with the logic outside. 

 So in this case really only to have a `const&` instead of just a `&`.

PR #165, opened 2023-11-21
Switch from push_back to variadic create in examples

tmadlener:
BEGINRELEASENOTES
- Make some of the examples slightly more idiomatic and avoid breaks from AIDASoft/podio#514

ENDRELEASENOTES


tmadlener:
Merging this as well since it doesn't change any functionality and is slightly more futureproof.

PR #167, opened 2023-11-26
Fix issue with narrowing

jmcarcell:
BEGINRELEASENOTES
- Fix issue with narrowing when compiling with clang; cast numbers to `float` which is the right type in EDM4hep instead of letting them being `int` or `double`

ENDRELEASENOTES

Needed after https://github.com/key4hep/k4FWCore/pull/165, maybe it will be fine when types are changed to double in EDM4hep.

Error message (when trying with doubles, originally it was int):

``` c++
/EDM4hep/edm4hep/edm4hep/MCParticleCollection.h:242:57: error: non-constant-expression cannot be narrowed from type 'double' to 'float' in initializer list [-Wc++11-narrowing]
  auto obj = new MCParticleObj({size, m_collectionID}, {std::forward<Args>(args)...});
                                                        ^~~~~~~~~~~~~~~~~~~~~~~~
/k4FWCore/test/k4FWCoreTest/src/components/ExampleFunctionalProducer.cpp:37:10: note: in instantiation of function template specialization 'edm4hep::MCParticleCollection::create<int, int, int, double, double, double>' requested here
    coll.create(1, 2, 3, 4., 5., 6.);
         ^
/EDM4hep/edm4hep/edm4hep/MCParticleCollection.h:242:57: note: insert an explicit cast to silence this issue
  auto obj = new MCParticleObj({size, m_collectionID}, {std::forward<Args>(args)...});
                                                        ^~~~~~~~~~~~~~~~~~~~~~~~
                                                        static_cast<float>(     ) 

```

### DIFF
--- a/test/k4FWCoreTest/src/components/ExampleFunctionalProducer.cpp
+++ b/test/k4FWCoreTest/src/components/ExampleFunctionalProducer.cpp
@@ -34,8 +34,8 @@ struct ExampleFunctionalProducer final : Gaudi::Functional::Producer<edm4hep::MC
   // This is the function that will be called to produce the data
   edm4hep::MCParticleCollection operator()() const override {
     auto coll = edm4hep::MCParticleCollection();
-    coll.create(1, 2, 3, 4, 5, 6);
-    coll.create(2, 3, 4, 5, 6, 7);
+    coll.create(1, 2, 3, static_cast<float>(4.), static_cast<float>(5.), static_cast<float>(6.));
+    coll.create(2, 3, 4, static_cast<float>(5.), static_cast<float>(6.), static_cast<float>(7.));
     // We have to return whatever collection type we specified in the
     // template argument
     return coll;


test/k4FWCoreTest/src/components/ExampleFunctionalProducer.cpp
```diff
@@ -34,8 +34,8 @@ struct ExampleFunctionalProducer final : Gaudi::Functional::Producer<edm4hep::MC
   // This is the function that will be called to produce the data
   edm4hep::MCParticleCollection operator()() const override {
     auto coll = edm4hep::MCParticleCollection();
-    coll.create(1, 2, 3, 4, 5, 6);
-    coll.create(2, 3, 4, 5, 6, 7);
+    coll.create(1, 2, 3, static_cast<float>(4.), static_cast<float>(5.), static_cast<float>(6.));
+    coll.create(2, 3, 4, static_cast<float>(5.), static_cast<float>(6.), static_cast<float>(7.));
```
tmadlener, line 38:
```suggestion
    coll.create(1, 2, 3, 4.f, 5.f, 6.f);
    coll.create(2, 3, 4, 5.f, 6.f, 7.f);
```
Would be slightly less noisy.

jmcarcell:
Yeah we should have one. In general more compilers should be shipped too so that it is easier to use them, right now clang is only built to have `clang-format`. Last time I tried to build with clang there were some issues with the tbb version being not compatible and Gaudi but I think that shouldn't be an issue anymore.

PR #171, opened 2023-12-08
Remove the legacy component that depend on the podio EventStore

tmadlener:
BEGINRELEASENOTES
- Remove the legacy components that depend on the podio EventStore. **This is a breaking change for anyone still using any of `PodioLegacyInput`, `PodioLegacyOutput` or `k4LegacyDataSvc`**.
  - If you still need the functionality to read old / lecagy input files, please let us know (e.g. via an issue).

ENDRELEASENOTES

I think now is a good time to do this since we have just made a release so that people can always go back to that in case this breaks things on the nightlies for them. We could then also figure out who is actually still using things legacy files and take things from there.

- [x] Check if this impacts other downstream packages


tmadlener:
It looks like we have switched over all our packages over to the non-legacy datasvc already. At least I could build a stack locally with these changes in without any further changes.

tmadlener:
As discussed at [todays meeting](https://indico.cern.ch/event/1356474/) we will merge this now and see if there are any complaints, resp. whether someone still needs this functionality.

PR #172, opened 2024-01-18
Don't use GaudiAlg

jmcarcell:
BEGINRELEASENOTES
- Don't use GaudiAlg since it's not built by default in Gaudi since https://gitlab.cern.ch/gaudi/Gaudi/-/merge_requests/1531

ENDRELEASENOTES

This can also serve as example for other repos, basically adding a `const EventContext&` in the arguments to `execute` and making `execute` const and changing `GaudiAlgorithm` to `Gaudi::Algorithm`. Headers change too:  `GaudiAlg/GaudiAlgorithm.h` to `Gaudi/Algorithm.h`


tmadlener (commented):
Looks good, but I am not sure we should keep all the "old" examples for creating example data, because I don't think our `DataHandle` is thread safe enough to just b marked `mutable`.

jmcarcell:
Alternative proposal (changes in the last commit): Remove the UniquePtr test which isn't doing anything than the other tests are not doing and set Cardinality to 1 which is how algorithms that can't run in parallel are set, for example: https://gitlab.cern.ch/gaudi/Gaudi/-/blob/master/GaudiKernel/include/GaudiKernel/Algorithm.h#L24

Having some algorithms around is good because not everything will be functional and it also allows to test if they can be mixed together.

jmcarcell:
If there aren't any more complaints I'll merge this today. I think the plan is to remove GaudiAlg for the version 40 of Gaudi so there is time but there are a few repos that have to be changed

tmadlener:
I think this looks good. One thing that is not entirely clear to me: Is the `EventContext` usable in any meaningful way for a "normal" algorithm? Or is that just used by the scheduler to make sure to use the right inputs?

jmcarcell:
I'm not sure I have seen that being used by an algorithm. There is an event number that could be used for seeding I guess but we have that information in the headers

PR #174, opened 2024-01-19
Make it possible to transparently migrate to TrackerHit interface

tmadlener:
BEGINRELEASENOTES
- Fix usages of `edm4hep::TrackerHit` to `edm4hep::TrackerHit3D` in order to keep things working after key4hep/EDM4hep#252
  - Check for the existance of the `TrackerHit3D` and tpyedef the `TrackerHit` to `TrackerHit3D` if it does not exist

ENDRELEASENOTES

- [x] Preprocessor ifs to have this working with older versions of EDM4hep for some time?
- [x] AIDASoft/podio#516
~- [x] key4hep/EDM4hep#252~

### DIFF
--- a/k4FWCore/components/PodioInput.cpp
+++ b/k4FWCore/components/PodioInput.cpp
@@ -95,6 +95,11 @@ void PodioInput::fillReaders() {
   m_readers["edm4hep::TrackerHit3DCollection"] = [&](std::string_view collName) {
     maybeRead<edm4hep::TrackerHit3DCollection>(collName);
   };
+  m_readers["edm4hep::TrackerHitCollection"] = [&](std::string_view collName) {
+    warning() << "Reading a TrackerHitCollection via TrackerHit3DCollection (" << collName
+              << "). Make sure to transition properly" << endmsg;
+    maybeRead<edm4hep::TrackerHit3DCollection>(collName);
+  };
   m_readers["edm4hep::TrackerHitPlaneCollection"] = [&](std::string_view collName) {
     maybeRead<edm4hep::TrackerHitPlaneCollection>(collName);
   };
--- a/k4FWCore/components/PodioInput.cpp
+++ b/k4FWCore/components/PodioInput.cpp
@@ -45,7 +45,14 @@
 #include "edm4hep/SimTrackerHitCollection.h"
 #include "edm4hep/TimeSeriesCollection.h"
 #include "edm4hep/TrackCollection.h"
+#if __has_include("edm4hep/TrackerHit3DCollection.h")
 #include "edm4hep/TrackerHit3DCollection.h"
+#else
+#include "edm4hep/TrackerHitCollection.h"
+namespace edm4hep {
+  using TrackerHit3DCollection = edm4hep::TrackerHitCollection;
+}  // namespace edm4hep
+#endif
 #include "edm4hep/TrackerHitPlaneCollection.h"
 #include "edm4hep/TrackerPulseCollection.h"
 #include "edm4hep/VertexCollection.h"
--- a/k4FWCore/components/PodioInput.cpp
+++ b/k4FWCore/components/PodioInput.cpp
@@ -45,7 +45,7 @@
 #include "edm4hep/SimTrackerHitCollection.h"
 #include "edm4hep/TimeSeriesCollection.h"
 #include "edm4hep/TrackCollection.h"
-#include "edm4hep/TrackerHitCollection.h"
+#include "edm4hep/TrackerHit3DCollection.h"
 #include "edm4hep/TrackerHitPlaneCollection.h"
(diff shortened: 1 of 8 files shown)


test/k4FWCoreTest/src/components/k4FWCoreTest_CreateExampleEventData.h
```diff
@@ -25,14 +25,24 @@
...
+// edm4hep
+#if __has_include("edm4hep/TrackerHit3DCollection.h")
+#include "edm4hep/TrackerHit3DCollection.h"
+#else
+#include "edm4hep/TrackerHitCollection.h"
+namespace edm4hep {
+  using TrackerHit3DCollection = edm4hep::TrackerHitCollection;
+}  // namespace edm4hep
+#endif
+#include "edm4hep/TrackerHitPlaneCollection.h"
```
jmcarcell, line 37:
```suggestion
```

test/k4FWCoreTest/src/components/ExampleFunctionalProducerMultiple.cpp
```diff
@@ -24,7 +24,15 @@
...
 #include "edm4hep/TrackCollection.h"
+#if __has_include("edm4hep/TrackerHit3DCollection.h")
+#include "edm4hep/TrackerHit3DCollection.h"
+#else
 #include "edm4hep/TrackerHitCollection.h"
+namespace edm4hep {
+  using TrackerHit3DCollection = edm4hep::TrackerHitCollection;
+}  // namespace edm4hep
+#endif
+#include "edm4hep/TrackerHitPlaneCollection.h"
```
jmcarcell, line 35:
```suggestion
```

test/k4FWCoreTest/src/components/ExampleFunctionalConsumerMultiple.cpp
```diff
@@ -23,7 +23,15 @@
...
 #include "edm4hep/TrackCollection.h"
+#if __has_include("edm4hep/TrackerHit3DCollection.h")
+#include "edm4hep/TrackerHit3DCollection.h"
+#else
 #include "edm4hep/TrackerHitCollection.h"
+namespace edm4hep {
+  using TrackerHit3DCollection = edm4hep::TrackerHitCollection;
+}  // namespace edm4hep
+#endif
+#include "edm4hep/TrackerHitPlaneCollection.h"
```
jmcarcell, line 34:
```suggestion
```
I don't think this is needed

k4FWCore/components/PodioInput.cpp
```diff
@@ -85,8 +92,13 @@ void PodioInput::fillReaders() {
   m_readers["edm4hep::ClusterCollection"] = [&](std::string_view collName) {
     maybeRead<edm4hep::ClusterCollection>(collName);
   };
+  m_readers["edm4hep::TrackerHit3DCollection"] = [&](std::string_view collName) {
+    maybeRead<edm4hep::TrackerHit3DCollection>(collName);
+  };
   m_readers["edm4hep::TrackerHitCollection"] = [&](std::string_view collName) {
-    maybeRead<edm4hep::TrackerHitCollection>(collName);
+    warning() << "Reading a TrackerHitCollection via TrackerHit3DCollection (" << collName
```
jmcarcell, line 99: I'm not sure about this warning; the only way you can reach here is when TrackerHitCollection is available (otherwise the type will never be `edm4hep::TrackerHitCollection`) in which case `TrackerHit3DCollection` is an alias to `TrackerHitCollection` meaning this warning wouldn't be true, no?

jmcarcell (approved):
After the comments looks fine to go in

k4FWCore/components/PodioInput.cpp
tmadlener: It will be possible to come here from old files, I think. In that case the old name will still be in the file metadata and propagate until here. However, we are fine just reading that through the new type. The warning might not be necessary though, as it will probably do the right thing. Additionally, all the other changes that we are currently introducing will break existing files in any case.

k4FWCore/components/PodioInput.cpp
```diff
@@ -85,8 +92,13 @@ void PodioInput::fillReaders() {
...
+              << "). Make sure to transition properly" << endmsg;
```
tmadlener, line 100:
```suggestion
```

k4FWCore/components/PodioInput.cpp
jmcarcell: How are old files going to be readable? I'm quite sure they are not. ROOT finds something called "TrackerHitCollection" that it doesn't know anything about and it's impossible for it to reconstruct a "Tracker3DHitCollection". So this is why I think this warning can't be triggered because if the collections are saved as `TrackerHitCollection` then they can't be read

PR #177, opened 2024-02-12
Format all python sources using black and add it to pre-commit

tmadlener:
BEGINRELEASENOTES
- Format all python sources using black and add it to pre-commit

ENDRELEASENOTES


tmadlener:
I can squash the non-formatting changes into one commit so that the format one can be more easily be put into the ignore-revs file

tmadlener:
Anyone with any objections for this?

jmcarcell:
I prefer `'` instead of `"` :cry: 

The choice should be the same for all the Key4hep repos with a single configuration file distributed everywhere. I guess it makes sense it's the black format since it's convenient to use it in `key4hep-spack`. So I don't have anything against this, once we agree I can do the formatting on all repos (not always easy depending on the number of open PRs...)

tmadlener:
> I prefer `'` instead of `"` 😢

Not much we can do there without post-black hooks, I am afraid. I would just stay with whatever it does.

For the other repos I would gradually go through. I also think this should be a bit less painful than e.g. c++ and clang-format, because there is not too much python code.

PR #179, opened 2024-02-22
Clean up the build workflows

jmcarcell:
BEGINRELEASENOTES
- Remove the builds for the release and nightlies from the "old" workflow since we already have a workflow that builds for all the supported OSes 

ENDRELEASENOTES

Builds with the LCG stack will fail until https://github.com/key4hep/k4FWCore/pull/172 is merged


tmadlener (approved):
will you also change the branch protection rules to require the new ones and remove the old ones?

jmcarcell:
Done

PR #180, opened 2024-02-22
Fix warnings related to the ROOTFrameReader and ROOTFrameWriter

jmcarcell:
BEGINRELEASENOTES
- Fix warnings related to the ROOTFrameReader and ROOTFrameWriter

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/components/PodioOutput.cpp
+++ b/k4FWCore/components/PodioOutput.cpp
@@ -38,7 +38,7 @@ StatusCode PodioOutput::initialize() {
     return StatusCode::FAILURE;
   }
 
-  m_framewriter = std::make_unique<podio::ROOTFrameWriter>(m_filename);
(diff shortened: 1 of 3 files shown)


tmadlener (approved):

jmcarcell:
Something looks broken on ubuntu :skull: 
```
ERROR: failed to load libk4FWCorePlugins.so: libTree.so.6.28: cannot open shared object file: No such file or directory
```

tmadlener:
I can reproduce that locally, but i have no idea why it fails. From the output it looks like all the paths and libraries are correct and also present (at least `libTree` is).

jmcarcell:
I'm not sure about the CI in ubuntu, I don't see anything wrong or think anything changed. So I'm going to merge this one, then retrigger a build a of the nightlies (without deploying) and will check there

PR #181, opened 2024-02-23
Fix deprecation warnings for changed mutable accessors

tmadlener:
BEGINRELEASENOTES
- Fix deprecation warnings for getting mutable references, introduce in AIDASoft/podio#553

ENDRELEASENOTES


jmcarcell:
I think the ubuntu workflows should be fixed with the nightly tomorrow

PR #184, opened 2024-02-29
Fix a few compiler warnings about unused status code

jmcarcell:
BEGINRELEASENOTES
- Fix a few compiler warnings about unused status code

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/include/k4FWCore/DataHandle.h
+++ b/k4FWCore/include/k4FWCore/DataHandle.h
@@ -90,9 +90,9 @@ template <typename T>
 DataHandle<T>::DataHandle(const std::string& descriptor, Gaudi::DataHandle::Mode a, IDataHandleHolder* fatherAlg)
     : DataObjectHandle<DataWrapper<T>>(descriptor, a, fatherAlg), m_eds("EventDataSvc", "DataHandle") {
   if (a == Gaudi::DataHandle::Writer) {
-    StatusCode sc [[maybe_unused]] = m_eds.retrieve();
-    m_dataPtr                      = nullptr;
-    auto* podio_data_service       = dynamic_cast<PodioDataSvc*>(m_eds.get());
+    StatusCode sc            = m_eds.retrieve().ignore();
+    m_dataPtr                = nullptr;
+    auto* podio_data_service = dynamic_cast<PodioDataSvc*>(m_eds.get());
     if (nullptr != podio_data_service) {
       if constexpr (std::is_integral_v<T> || std::is_floating_point_v<T>) {
         m_dataPtr = new T();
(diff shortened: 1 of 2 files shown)


k4FWCore/include/k4FWCore/DataHandle.h
```diff
@@ -90,9 +90,9 @@ template <typename T>
 DataHandle<T>::DataHandle(const std::string& descriptor, Gaudi::DataHandle::Mode a, IDataHandleHolder* fatherAlg)
     : DataObjectHandle<DataWrapper<T>>(descriptor, a, fatherAlg), m_eds("EventDataSvc", "DataHandle") {
   if (a == Gaudi::DataHandle::Writer) {
-    StatusCode sc            = m_eds.retrieve();
-    m_dataPtr                = nullptr;
-    auto* podio_data_service = dynamic_cast<PodioDataSvc*>(m_eds.get());
+    StatusCode sc [[maybe_unused]] = m_eds.retrieve();
```
andresailer, line 93: a) Why would we be ignoring this status code?

b) Doesn't this exist?

```suggestion
   m_eds.retrieve().ignore();
```

jmcarcell: I think that's because you get an error when trying to retrieve something that doesn't exist. 
I changed it to use `ignore()`

jmcarcell: The nullptr check was introduced there to be able to use the LegacyDataSvc in #117, that can be removed now. I'll add that to this PR since it's some sort of cleanup

jmcarcell: The nullptr check seems to be there for when another data service could be used, maybe from when there was the legacy one. I think probably the nullptr check and the check below can be removed

andresailer: But we need some way to report that something wasn't found or is this handled somewhere else?

jmcarcell: If something is not found in the store then that's handled by https://github.com/key4hep/k4FWCore/main/no-datasvc/k4FWCore/include/k4FWCore/DataHandle.h#L116 when using our DataHandles, if using the ones from Gaudi then it's handled by Gaudi

andresailer: Link gives "Not Found"

jmcarcell: Now it's correct, I got it from emacs from the branch I was using and then I have to edit the link and put `main` in the right place if I don't want to link to another branch and I put it in the wrong place

PR #185, opened 2024-03-18
Use setMomentum for particles

jmcarcell:
BEGINRELEASENOTES
- Use setMomentum for particles, fixing an issue when trying to change a const reference

ENDRELEASENOTES

I'm not sure what has changed since I think that has been there for a time but
the reference from `getMomentum` should be const. Locally I don't need this fix
with clang 17 and I can't reproduce the error when building on top of the latest
nightlies (that are picking up EDM4hep from a few days ago) but on spack I can
reproduce consistently


tmadlener:
What has changed is this: https://github.com/AIDASoft/podio/pull/553 (which is not part of podio v00-99)

tmadlener (approved):

PR #186, opened 2024-03-22
Handle SIGPIPE cleanly

kjvbrt:
BEGINRELEASENOTES
- Handle SIGPIPE signal

ENDRELEASENOTES

Cleanly terminates k4run when one uses it with `head`. Example:
```
k4run --dry-run pythia.py | head -n 1
[k4run] --> Pythia8 --> HepMCToEDMConverter --> StableParticles --> out
```
[more details about Python and SIGPIPE](https://docs.python.org/3/library/signal.html#note-on-sigpipe)

Triggered by https://github.com/HEP-FCC/fcc-tutorials/pull/140


jmcarcell (approved):
Looks good. Using head and maybe grep when it finds a single match and exits early

Issue #187, opened 2024-03-24
EventDataSvc not found on macOS during build


veprbl:
- OS version: macOS 10.14
 - Compiler version: clang 11
 - Package version: any recent version, including v01-00pre19, Gaudi 36r16
 - Reproduced by: try to build
 - Input: N/A
 - Output: full build and run log output

```output
cd /source/build/k4FWCore && ../run /bin/listcomponents --output k4FWCorePlugins.components libk4FWCorePlugins.so
cd /source/build/k4FWCore && ../run /bin/genconf --configurable-module=GaudiKernel.Proxy --configurable-default-name=Configurable.DefaultName --configurable-algorithm=ConfigurableAlgorithm --configurable-algtool=ConfigurableAlgTool --configurable-auditor=ConfigurableAuditor --configurable-service=ConfigurableService -o genConfDir/k4FWCore -p k4FWCore -i /source/build/k4FWCore/libk4FWCorePlugins.so
[info   ] :::::: libraries : [ /source/build/k4FWCore/libk4FWCorePlugins.so ] ::::::
[info   ] :::: processing library: /source/build/k4FWCore/libk4FWCorePlugins.so...
[info   ]  - component: EventHeaderCreator (Algorithm)
[info   ]  - component: FCCDataSvc (Service)
[info   ]  - component: PodioInput (Algorithm)
ERROR ServiceLocatorHelper::service: can not locate service EventDataSvc
[error  ] Error instantiating PodioInput from /source/build/k4FWCore/libk4FWCorePlugins.so
[error  ] Got exception: Service [EventDataSvc] not found
[info   ]  - component: PodioOutput (Algorithm)
[info   ]  - component: UniqueIDGenSvc (Service)
[info   ]  - component: k4DataSvc (Service)
[info   ] :::::: libraries : [ /source/build/k4FWCore/libk4FWCorePlugins.so ] :::::: [DONE]
make[2]: *** [k4FWCore/CMakeFiles/k4FWCorePlugins.dir/build.make:205: k4FWCore/libk4FWCorePlugins.so] Error 1
```

The exception is thrown by the `evtSvc()` in the `PodioInput::PodioInput(...)` constructor.

 - Goal: build the software

Somehow, there is no such issue on Linux with all the same versions (except built with gcc, maybe)

PR #188, opened 2024-03-24
PodioInput: don't do anything in constructor during the genconf step

veprbl:
BEGINRELEASENOTES
- Fix a bug that was preventing successful build on macOS by not accessing EventDataSvc in the constructor of the `PodioInput`.

ENDRELEASENOTES

Fixes: #187


jmcarcell (approved):
I have been annoyed at this for a while but never fixed it since on Linux all you get is
```
[424/487] Linking CXX shared module _d...e-build/k4FWCore/libk4FWCorePlugins.so
DefaultName         ERROR Could not get PodioDataSvc
```

PR #194, opened 2024-05-10
Make the Lcio2EDM4hepTool a friend of PodioDataSvc 

Zehvogel:
BEGINRELEASENOTES
- Make the Lcio2EDM4hepTool a friend of PodioDataSvc to enable event parameter conversion

ENDRELEASENOTES

See discussion in: https://github.com/key4hep/k4MarlinWrapper/pull/184

### DIFF
--- a/k4FWCore/include/k4FWCore/PodioDataSvc.h
+++ b/k4FWCore/include/k4FWCore/PodioDataSvc.h
@@ -50,6 +50,7 @@ template <typename T> class MetaDataHandle;
 class PodioDataSvc : public DataSvc {
   template <typename T> friend class MetaDataHandle;
   friend class PodioOutput;
+  friend class Lcio2EDM4hepTool;
 
 public:
   typedef std::vector<std::pair<std::string, podio::CollectionBase*>> CollRegistry;


k4FWCore/include/k4FWCore/PodioDataSvc.h
```diff
@@ -93,7 +93,7 @@ class PodioDataSvc : public DataSvc {
     return DataSvc::registerObject("/Event", "/" + collName, wrapper);
   }
 
-  const podio::Frame& getEventFrame() const { return m_eventframe; }
+  podio::Frame& getEventFrame() { return m_eventframe; }
```
m-fila, line 96: How about having both overloads for const-correctness?
```suggestion
  podio::Frame& getEventFrame() { return m_eventframe; }
  const podio::Frame& getEventFrame() const { return m_eventframe; }
```

tmadlener: Or potentially having a dedicated `getMutableEventFrame` (plus documentation that this might go away again)?

andresailer: Or make MarlinWrapper a friend to give access to protected functions.

Zehvogel: > Or make MarlinWrapper a friend to give access to protected functions.

This might be my favourite solution so far as it actively stops others from perpetrating the same kind of abuse

m-fila: In that case, what do you think about using passkey idiom to limit the scope of friendship?

andresailer: Given that we fully control MarlinWrapper, I would follow the KISS idiom.

andresailer (approved):

andresailer (approved):

PR #195, opened 2024-05-16
Make MetaDataHandle::get throw in case the value is not available and add more overloads

tmadlener:
BEGINRELEASENOTES
- Make `MetaDataHandle::get` throw an exception in case the value for the handle is not (yet) available.
- Add `MetaDataHandle::get(T defaultValue)` overload to get a default value and no exception in case the value is not (yet) available.
- Add `MetaDataHandle::get_optional()` to retrieve an optional that is engaged and holds the value if the handle already has a value available.
  - The optional is returned directly from the underlying Frame in case that is available (see  [AIDASoft/podio#580](https://github.com/AIDASoft/podio/pull/580)) or is constructed in place in case it is not yet available.
 
ENDRELEASENOTES

~I have for now opted to let the underlying change in podio shine through here. I think having an `optional` value for `MetaDataHandle::get` offers the same benefits as having it on the `Frame` in the first place, as otherwise there is no way of distinguishing between empty and unset values.~

### DIFF
--- a/k4FWCore/include/k4FWCore/MetaDataHandle.h
+++ b/k4FWCore/include/k4FWCore/MetaDataHandle.h
@@ -28,12 +28,22 @@ public:
   MetaDataHandle(const Gaudi::DataHandle& handle, const std::string& descriptor, Gaudi::DataHandle::Mode a);
   ~MetaDataHandle();
 
+  /// Get the (optional) value that is stored in this MetaDataHandle
+  ///
+  /// @returns A std::optional that is engaged if the value was available as
+  ///          metadata
   const auto get() const;
-  void       put(T);
+
+  /// Set the value for this MetaDataHandle
+  ///
+  /// @note This can only be called during initalize and/or finalize but not
+  /// during execute for algorithms that use it
+  void put(T);
 
 private:
   std::string fullDescriptor() const;
-  void        checkPodioDataSvc();
+
+  void checkPodioDataSvc();
 
 private:
   ServiceHandle<IDataProviderSvc> m_eds;
--- a/k4FWCore/include/k4FWCore/MetaDataHandle.h
+++ b/k4FWCore/include/k4FWCore/MetaDataHandle.h
@@ -30,8 +30,14 @@ public:
 
   /// Get the (optional) value that is stored in this MetaDataHandle
   ///
-  /// @returns A std::optional that is engaged if the value was available as
-  ///          metadata
+  /// @note The return type of this changes depending on the podio version that
+  /// has been used to build k4FWCore. For pre 1.0 versions of podio this will
+  /// return a default initialized (empty) value in case the underlying
+  /// parameter could not obtained be. For later podio versions this will return
+  /// a std::optional.
+  ///
+  /// @returns Either the (potentially default initialized) value or a
+  /// std::optional that is engaged in case the value is available as metadata
   const auto get() const;
 
   /// Set the value for this MetaDataHandle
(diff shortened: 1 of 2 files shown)


tmadlener:
It looks like this has a rather small impact on downstream code, so I would be in favor of having this as a breaking change, i.e. the `std::optional` appearing in the return type.

jmcarcell:
I'm fine with that, k4FWCore doesn't build anyway on the latest release, so not much is going to change

k4FWCore/include/k4FWCore/MetaDataHandle.h
```diff
@@ -73,7 +65,7 @@ MetaDataHandle<T>::MetaDataHandle(const Gaudi::DataHandle& handle, const std::st
 }
 
 //---------------------------------------------------------------------------
```
andresailer, line 67: Add docstring

tmadlener: I have added a docstring that only mentions `std::optional`. Should it also mention that the return type depends on the version of podio? I am not entirely sure whether that does not confusing.

andresailer: Mentioning that this depends on podio version would be good, because people can see the std::optional or T in doxygen, but they can't see that this might change.

andresailer: Add the possibility to add a default value, and if no default value is given throw the exception.

tmadlener:
Now I have to introduce pre processor checks here and check that it works for podio @0.99 as well as once #580 is merged. I will be coming back to this a bit later (i.e. probably tomorrow).

k4FWCore/include/k4FWCore/MetaDataHandle.h
m-fila: How about still having a version returning optional e.g.`get_optional`? I think it's pretty common to have throwing, default and optional getters for different cases. This way we don't mess-up with the return here but still provide a way to handle missing values when an exception is too much

tmadlener: `get_optional` is done and the others are now using that and simply changing return values if necessary. Additionally, `get_optional` is now also the single place where all the podio version differences are handled and a `std::optional` is created on the fly with the heuristic check for empty values to return an empty optional.

tmadlener:
Anything still missing from here, or can we merge this?

test/k4FWCoreTest/src/components/k4FWCoreTest_cellID_reader.cpp
```diff
@@ -41,10 +43,10 @@ StatusCode k4FWCoreTest_cellID_reader::initialize() {
 StatusCode k4FWCoreTest_cellID_reader::execute(const EventContext&) const {
   const auto simtrackerhits_coll = m_simTrackerHitReaderHandle.get();
 
-  auto       collID    = simtrackerhits_coll->getID();
-  const auto cellIDstr = m_cellIDHandle.get();
+  const auto cellIDstr = m_cellIDHandle.get("");
```
tmadlener, line 46:
```suggestion
  const auto cellIDstr = m_cellIDHandle.get();
```

This allows us to differentiate between error in retrieval, and retrieval of wrong value.

test/k4FWCoreTest/src/components/k4FWCoreTest_cellID_reader.cpp
```diff
@@ -19,6 +19,8 @@
 #include "k4FWCoreTest_cellID_reader.h"
 #include "k4FWCoreTest_cellID_writer.h"
 
+#include "podio/podioVersion.h"
+
```
tmadlener, line 23:
```suggestion
```

tmadlener:
Merging this, since this is not transparent and we need to unblock AIDASoft/podio#580

PR #198, opened 2024-05-28
Remove the dQ/dx information from the tracks

jmcarcell:
BEGINRELEASENOTES
- Remove the dQ/dx information from the tracks. Needed after https://github.com/key4hep/EDM4hep/pull/311

ENDRELEASENOTES

Doesn't have any effect downstream since they were used only for tests


jmcarcell:
Merging this since it's uncontroversial and it will be ready in the next nightlies

PR #199, opened 2024-06-05
Remove header that has become unnecessary

tmadlener:
BEGINRELEASENOTES
- Remove the `rootUtils.h` header that has been copied from podio since it has become obsolete with #171 

ENDRELEASENOTES


jmcarcell (approved):
I get this `rootUtils.h` highlighted all the time in my editor as "useless"

PR #200, opened 2024-06-17
Use `edm4hep::labels`

jmcarcell:
BEGINRELEASENOTES
- Use `edm4hep::labels`

ENDRELEASENOTES


tmadlener (approved):
Looks good to me since we are in any case no longer backwards compatible with EDM4hep <= 0.10.5

PR #201, opened 2024-06-18
Use `std::vector` instead of `std::map` for reading or writing an arbitrary number of collections.

jmcarcell:
The current implementation with std::map is not easy to use. The collections get ordered alphabetically in std::map so the order set in the steering file is lost, which means if there is some kind of mapping (like one output collection for each input collection) then one has to reproduce this map in the algorithm and the way I found to do it is to have another property with the input and output names again. For output one creates the collections and then puts them in a vector, I think it can't get simpler than that. For input, since there can't be vectors to references one gets a pointer to the collection, which is not ideal but not so bad. With a custom type it would be possible to have a vector that when using [] can get a reference but then that introduces a custom type which I don't like.

BEGINRELEASENOTES

- FunctionalAlgorithms: Use std::vector instead of std::map for input or output an arbitrary number of collections.
- Add an `inputLocations` and `outputLocations` methods to be able to retrieve the locations set in the steering file.

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/include/k4FWCore/Consumer.h
+++ b/k4FWCore/include/k4FWCore/Consumer.h
@@ -41,8 +41,8 @@ namespace k4FWCore {
         : Gaudi::Functional::details::DataHandleMixin<std::tuple<>, std::tuple<>, Traits_> {
       using Gaudi::Functional::details::DataHandleMixin<std::tuple<>, std::tuple<>, Traits_>::DataHandleMixin;
 
-      static_assert(((std::is_base_of_v<podio::CollectionBase, In> || isMapToCollLike<In>::value) && ...),
-                    "Consumer input types must be EDM4hep collections or maps to collections");
+      static_assert(((std::is_base_of_v<podio::CollectionBase, In> || isVectorLike_v<In>) && ...),
+                    "Consumer input types must be EDM4hep collections or vectors to collection pointers");
 
       template <typename T>
       using InputHandle_t = Gaudi::Functional::details::InputHandle_t<Traits_, std::remove_pointer_t<T>>;
@@ -56,7 +56,7 @@ namespace k4FWCore {
 
       template <typename IArgs, std::size_t... I>
       Consumer(std::string name, ISvcLocator* locator, const IArgs& inputs, std::index_sequence<I...>)
-          : base_class(std::move(name), locator),
+        : base_class(std::move(name), locator),
             // The input locations are filled by creating a property with a callback function
             // that creates the handles because that is when the input locations become available
             // (from a steering file, for example) and the handles have to be created for
@@ -71,7 +71,8 @@ namespace k4FWCore {
                   }
                   std::get<I>(m_inputs) = std::move(handles);
                 },
-                Gaudi::Details::Property::ImmediatelyInvokeHandler{true}}...} {}
+                Gaudi::Details::Property::ImmediatelyInvokeHandler{true}}...}
+      {}
 
       Consumer(std::string name, ISvcLocator* locator,
                Gaudi::Functional::details::RepeatValues_<KeyValues, sizeof...(In)> const& inputs)
--- a/k4FWCore/include/k4FWCore/Consumer.h
+++ b/k4FWCore/include/k4FWCore/Consumer.h
@@ -92,7 +92,7 @@ namespace k4FWCore {
       // ... instead, they must implement the following operator
       virtual void operator()(const In&...) const = 0;
 
-      const auto inputLocations(int i) const {
+      auto inputLocations(int i) const {
         if (i >= sizeof...(In)) {
           throw GaudiException("Called inputLocations with an index out of range", "Consumer", StatusCode::FAILURE);
         }
--- a/k4FWCore/include/k4FWCore/Consumer.h
+++ b/k4FWCore/include/k4FWCore/Consumer.h
@@ -27,6 +27,7 @@
 
 #include "k4FWCore/FunctionalUtils.h"
 
+#include <ranges>
 #include <type_traits>
 #include <utility>
 
@@ -90,6 +91,13 @@ namespace k4FWCore {
 
       // ... instead, they must implement the following operator
       virtual void operator()(const In&...) const = 0;
+
+      const auto inputLocations(int i) const {
+        if (i >= sizeof...(In)) {
+          throw GaudiException("Called inputLocations with an index out of range", "Consumer", StatusCode::FAILURE);
+        }
+        return m_inputLocations[i] | std::views::transform([](const DataObjID& id) -> const auto& { return id.key(); });
+      }
     };
 
   }  // namespace details
--- a/k4FWCore/include/k4FWCore/Consumer.h
+++ b/k4FWCore/include/k4FWCore/Consumer.h
@@ -28,6 +28,7 @@
 #include "k4FWCore/FunctionalUtils.h"
 
 #include <ranges>
+#include <stdexcept>
 #include <type_traits>
 #include <utility>
 
@@ -42,7 +43,7 @@ namespace k4FWCore {
         : Gaudi::Functional::details::DataHandleMixin<std::tuple<>, std::tuple<>, Traits_> {
       using Gaudi::Functional::details::DataHandleMixin<std::tuple<>, std::tuple<>, Traits_>::DataHandleMixin;
 
-      static_assert(((std::is_base_of_v<podio::CollectionBase, In> || isVectorLike_v<In>)&&...),
+      static_assert(((std::is_base_of_v<podio::CollectionBase, In> || isVectorLike_v<In>) && ...),
                     "Consumer input types must be EDM4hep collections or vectors of collection pointers");
 
       template <typename T>
@@ -92,11 +93,29 @@ namespace k4FWCore {
       // ... instead, they must implement the following operator
       virtual void operator()(const In&...) const = 0;
 
-      auto inputLocations(int i) const {
+      /**
+       * @brief    Get the input locations for a given input index
+       * @param i  The index of the input
+       * @return   A range of the input locations
+       */
+      const auto inputLocations(int i) const {
         if (i >= sizeof...(In)) {
-          throw GaudiException("Called inputLocations with an index out of range", "Consumer", StatusCode::FAILURE);
+          throw std::out_of_range("Called inputLocations with an index out of range");
         }
         return m_inputLocations[i] | std::views::transform([](const DataObjID& id) -> const auto& { return id.key(); });
+      }
+      /**
+       * @brief       Get the input locations for a given input name
+       * @param name  The name of the input
+       * @return      A range of the input locations
+       */
+      const auto inputLocations(std::string_view name) const {
+        auto it = std::find_if(m_inputLocations.begin(), m_inputLocations.end(),
+                               [&name](const auto& prop) { return prop.name() == name; });
+        if (it == m_inputLocations.end()) {
+          throw std::runtime_error("Called inputLocations with an unknown name");
+        }
+        return it->value() | std::views::transform([](const DataObjID& id) -> const auto& { return id.key(); });
       }
     };
 
--- a/k4FWCore/include/k4FWCore/Consumer.h
+++ b/k4FWCore/include/k4FWCore/Consumer.h
@@ -43,7 +43,7 @@ namespace k4FWCore {
         : Gaudi::Functional::details::DataHandleMixin<std::tuple<>, std::tuple<>, Traits_> {
       using Gaudi::Functional::details::DataHandleMixin<std::tuple<>, std::tuple<>, Traits_>::DataHandleMixin;
 
-      static_assert(((std::is_base_of_v<podio::CollectionBase, In> || isVectorLike_v<In>)&& ...),
+      static_assert(((std::is_base_of_v<podio::CollectionBase, In> || isVectorLike_v<In>)&&...),
                     "Consumer input types must be EDM4hep collections or vectors of collection pointers");
 
       template <typename T>
@@ -100,7 +100,8 @@ namespace k4FWCore {
        */
       const auto inputLocations(int i) const {
         if (i >= sizeof...(In)) {
-          throw std::out_of_range("Called inputLocations with an index out of range");
+          throw std::out_of_range("Called inputLocations with an index out of range, index: " + std::to_string(i) +
+                                  ", number of inputs: " + std::to_string(sizeof...(In)));
         }
         return m_inputLocations[i] | std::views::transform([](const DataObjID& id) -> const auto& { return id.key(); });
       }
@@ -110,8 +111,7 @@ namespace k4FWCore {
        * @return      A range of the input locations
        */
       const auto inputLocations(std::string_view name) const {
-        auto it = std::find_if(m_inputLocations.begin(), m_inputLocations.end(),
-                               [&name](const auto& prop) { return prop.name() == name; });
+        auto it = std::ranges::find_if(m_inputLocations, [&name](const auto& prop) { return prop.name() == name; });
         if (it == m_inputLocations.end()) {
           throw std::runtime_error("Called inputLocations with an unknown name");
         }
--- a/k4FWCore/include/k4FWCore/Consumer.h
+++ b/k4FWCore/include/k4FWCore/Consumer.h
@@ -41,7 +41,7 @@ namespace k4FWCore {
         : Gaudi::Functional::details::DataHandleMixin<std::tuple<>, std::tuple<>, Traits_> {
       using Gaudi::Functional::details::DataHandleMixin<std::tuple<>, std::tuple<>, Traits_>::DataHandleMixin;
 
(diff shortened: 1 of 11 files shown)


m-fila (commented):
I had a quick look. There are still a few occurrences  of 'map' where 'vector' should be used

Functions `detail::readMapInputs` and `detail::putMapOutputs` could also be renamed since they operate on vectors

I think `const *` are acceptable

In case the information about collection keys will be needed:
In Gaudi there is`inputLocation()` and `outputLocation()` (defined in `DataHandleMixing`) that can be used to get keys of inputs and outputs (like tuple either by index or type if unique). From what I seen they are not accessible from k4FWCore functional algorithms. It might be good idea to add something similar here. 
Alternatively something like `std::get<n>(m_outputs).at(m).fullKey().key()` could be used to get key of `n`-th output and `m`-th collection (given n-th output is a vector of collections)

jmcarcell:
> I had a quick look. There are still a few occurrences of 'map' where 'vector' should be used
> 
> Functions `detail::readMapInputs` and `detail::putMapOutputs` could also be renamed since they operate on vectors
> 
> I think `const *` are acceptable
> 
> In case the information about collection keys will be needed: In Gaudi there is`inputLocation()` and `outputLocation()` (defined in `DataHandleMixing`) that can be used to get keys of inputs and outputs (like tuple either by index or type if unique). From what I seen they are not accessible from k4FWCore functional algorithms. It might be good idea to add something similar here. Alternatively something like `std::get<n>(m_outputs).at(m).fullKey().key()` could be used to get key of `n`-th output and `m`-th collection (given n-th output is a vector of collections)

There is also `m_inputLocations` and `m_outputLocations` that can be used like here: https://github.com/key4hep/k4Reco/pull/2/files#diff-fd32a841abe95600867a31e1e8a8f06232c5fe7e654319fdfad0d14b55760bc0R256. I don't have a strong opinion on implementing `inputLocation()` and `outputLocation()` or not.

jmcarcell:
I added a `inputLocations` and `outputLocations` to retrieve those and they are used in a couple of tests. Now I notice the Gaudi version is in singular but here it can be more than one. I think them being views is fine, but they can easily be changed to a vector.

k4FWCore/include/k4FWCore/Consumer.h
```diff
@@ -90,6 +91,13 @@ namespace k4FWCore {
 
       // ... instead, they must implement the following operator
       virtual void operator()(const In&...) const = 0;
+
+      auto inputLocations(int i) const {
+        if (i >= sizeof...(In)) {
+          throw GaudiException("Called inputLocations with an index out of range", "Consumer", StatusCode::FAILURE);
+        }
+        return m_inputLocations[i] | std::views::transform([](const DataObjID& id) -> const auto& { return id.key(); });
```
tmadlener, line 99: This will require a change here:

https://github.com/key4hep/k4FWCore/blob/beb9affb476f97f313a6d0ba20f914ea86100a61/CMakeLists.txt#L42-L46

jmcarcell: Indeed, actually this should have been changed earlier. For spack it's fine and locally I set it to C++20, if no one complained, is no one building k4FWCore? ...

tmadlener: I am, but also with c++20 :sweat_smile:

jmcarcell:
Any more comments or comments about the functions to get the locations? @m-fila

m-fila:
> Any more comments or comments about the functions to get the locations? @m-fila

Sorry I was busy with other things, I'll take another look

test/k4FWCoreTest/src/components/ExampleFunctionalConsumerRuntimeCollections.cpp
```diff
@@ -26,20 +26,26 @@
...
   // This is the function that will be called to produce the data
-  void operator()(const std::map<std::string, const edm4hep::MCParticleCollection&>& input) const override {
+  void operator()(const std::vector<const edm4hep::MCParticleCollection*>& input) const override {
     if (input.size() != 3) {
-      throw std::runtime_error("Wrong size of the input map, expected 3, got " + std::to_string(input.size()));
+      throw std::runtime_error("Wrong size of the input vector, expected 3, got " + std::to_string(input.size()));
     }
-    for (auto& [key, val] : input) {
+    for (int i = 0; i < 3; i++) {
+      if (inputLocations(0)[i] != "MCParticles" + std::to_string(i)) {
```
tmadlener, line 41: There is no easy way to have `inputLocations` take the key from the `KeyValues` above, right? I.e. such that something like this would becom possible

```cpp
if (inputLocations("InputCollection")[i] == "MCParticles" + std::to_string(i)) { /**/ }
```

It doesn't really matter here, but if there are multiple input variable length inputs, this would make it much harder to break things, I think. However, I am not sure how often that actually is the case in real world uses.

m-fila: There is:
- either by property (returning `std::vector<std::string>>`)
  ```cpp
  auto inputLocations(const std::string& name) const {
    Gaudi::Property<std::vector<std::string>> tmp;
    tmp.assign(this->getProperty(name));
    return tmp.value();
  }
  ```
 - or from `m_inputlocations` (returing view)
  ```cpp
  auto intpuLocations(const std::string& name) const {
    auto it = std::ranges::find_if(m_inputLocations, [&name](const auto& prop) { return prop.name() == name; });
    if (it == m_inputLocations.end()) {
      throw GaudiException("Input named " + name + " not found", "Consumer", StatusCode::FAILURE);
    }
    return *it | std::views::transform([](const DataObjID& id) -> const auto& { return id.key(); });
  }
  ```

It could be also nice to get by type (if it's unique, there is a tuple underneath) but that seems to be more complicated as `detail::transformType` transforms all the collection types to `podio::CollectionBase`

k4FWCore/include/k4FWCore/Transformer.h
```diff
@@ -126,6 +125,17 @@ namespace k4FWCore {
         }
       }
 
+      auto inputLocations(int i) const {
+        if (i >= sizeof...(In)) {
+          throw GaudiException("Called inputLocations with an index out of range", "Consumer", StatusCode::FAILURE);
+        }
+        return m_inputLocations[i] | std::views::transform([](const DataObjID& id) -> const auto& { return id.key(); });
+      }
```
m-fila, line 133: The check could happen at compile time since the number of input/output arguments is known. I'm not sure if there is a point in allowing a runtime argument to this. If there is then probably then should be a function to get the number of inputs/outputs (Gaudi functional algorithms have `InputLocationSize` to get `sizeof...(In)`)
```suggestion
      template<size_t i>
      auto inputLocations() const {
        static_assert(i < sizeof...(In), "Index must be smaller than the number of inputs of the algorithm");
        return m_inputLocations[i] | std::views::transform([](const DataObjID& id) -> const auto& { return id.key(); });
      }
    };
```

k4FWCore/include/k4FWCore/Consumer.h
```diff
@@ -90,6 +91,13 @@ namespace k4FWCore {
 
       // ... instead, they must implement the following operator
       virtual void operator()(const In&...) const = 0;
+
+      auto inputLocations(int i) const {
```
m-fila, line 95: Since this function is supposed to be used by algorithm developers a docstring could be useful

k4FWCore/include/k4FWCore/Transformer.h
```diff
@@ -126,6 +125,17 @@ namespace k4FWCore {
         }
       }
 
+      auto inputLocations(int i) const {
+        if (i >= sizeof...(In)) {
+          throw GaudiException("Called inputLocations with an index out of range", "Consumer", StatusCode::FAILURE);
```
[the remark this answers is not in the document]
jmcarcell: In general I think all of these should go, I don't like that there is a tag that can't be removed in `GaudiException`; the name of the algorithm will always appear on the left and then the message is "TAG : ERROR_MSG" where here the tag is Consumer which I don't think is needed.

k4FWCore/include/k4FWCore/Transformer.h
jmcarcell: But the index is not known at compile time, for example, when running inside a loop or any other logic that someone may use for indexing so it's not possible to do a `static_assert`. For the other comment I don't think we need a function to call `sizeof...` which is short enough and there is not a lot of use case for that from an algorithm given that it is known, no?

k4FWCore/include/k4FWCore/Consumer.h
jmcarcell: Docstring added for all the new location functions.

test/k4FWCoreTest/src/components/ExampleFunctionalConsumerRuntimeCollections.cpp
jmcarcell: Added functions to get by name. Getting by type is not trivial since one still has to get the index to find where the vector with the names is... so maybe if someone asks for it? By index and name should be plenty of options.

tmadlener: I think that the use case for getting this by type is limited(?). i would assume that in that case you don't really care about the collection name since it's only one in any case. The only reason I could think of would be for log messages, but in that case, one of the overloads here with index or name, should also work, right?

k4FWCore/include/k4FWCore/Transformer.h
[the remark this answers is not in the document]
m-fila: I completely agree. Maybe it was for historical reasons. I think the standard exceptions that you have now are nicer :+1:

k4FWCore/include/k4FWCore/Consumer.h
```diff
@@ -90,6 +92,31 @@ namespace k4FWCore {
...
       virtual void operator()(const In&...) const = 0;
+
+      /**
+       * @brief    Get the input locations for a given input index
+       * @param i  The index of the input
+       * @return   A range of the input locations
+       */
+      const auto inputLocations(int i) const {
+        if (i >= sizeof...(In)) {
+          throw std::out_of_range("Called inputLocations with an index out of range");
```
tmadlener, line 103: Maybe add how many would be available in the message?

k4FWCore/include/k4FWCore/Consumer.h
```diff
@@ -90,6 +92,31 @@ namespace k4FWCore {
...
+        }
+        return m_inputLocations[i] | std::views::transform([](const DataObjID& id) -> const auto& { return id.key(); });
+      }
+      /**
+       * @brief       Get the input locations for a given input name
+       * @param name  The name of the input
+       * @return      A range of the input locations
+       */
+      const auto inputLocations(std::string_view name) const {
+        auto it = std::find_if(m_inputLocations.begin(), m_inputLocations.end(),
```
tmadlener, line 113:
```suggestion
        auto it = std::ranges::find_if(m_inputLocations,
```
Since we can use c++20 now.

k4FWCore/include/k4FWCore/Transformer.h
```diff
@@ -126,6 +126,51 @@ namespace k4FWCore {
...
+        }
+        return m_inputLocations[i] | std::views::transform([](const DataObjID& id) -> const auto& { return id.key(); });
+      }
+      /**
+       * @brief       Get the input locations for a given input name
+       * @param name  The name of the input
+       * @return      A range of the input locations
+       */
+      const auto inputLocations(std::string_view name) const {
+        auto it = std::find_if(m_inputLocations.begin(), m_inputLocations.end(),
```
tmadlener, line 146:
```suggestion
        auto it = std::ranges::find_if(m_inputLocations,
```

k4FWCore/include/k4FWCore/Transformer.h
```diff
@@ -202,14 +247,65 @@ namespace k4FWCore {
...
```
tmadlener, line 275:
```suggestion
        auto it = std::ranges::find_if(m_inputLocations,
```

k4FWCore/include/k4FWCore/Transformer.h
```diff
@@ -202,14 +247,65 @@ namespace k4FWCore {
...
+        return m_outputLocations[i] |
+               std::views::transform([](const DataObjID& id) -> const auto& { return id.key(); });
+      }
+      /**
+       * @brief       Get the output locations for a given output name
+       * @param name  The name of the output
+       * @return      A range of the output locations
+       */
+      const auto outputLocations(std::string_view name) const {
+        auto it = std::find_if(m_outputLocations.begin(), m_outputLocations.end(),
```
tmadlener, line 301:
```suggestion
        auto it = std::ranges::find_if(m_outputLocations,
```

k4FWCore/include/k4FWCore/Transformer.h
m-fila: We could demand the index at the compile time (template argument) if we don't want people to write loops over arguments. If someone wrote
```cpp
vector<MCParticleCollection> operator()(const vector<const MCParticleCollection*>&,  const vector<const SimTrackerHitCollection*>&)
```
 then I'd expect they should be able to understand "at compile time" that `vector<MCParticleCollection*>` is the first argument  and `vector<SimTrackerHitCollection*>` is the second argument. Using an index is just a convention

If writing loops over arguments is allowed then I think we should provide some way to get the bound. `sizeof...(In)` looks to me a bit too much of an implementation detail, since the algorithm developers probably don't have to use `In` directly at all. 
```cpp
for(size_t i = 0; i != sizeof...(In); ++i){
  auto locations = inputLocations(i);
}
```
looks a bit more daunting than
```cpp
for(size_t i = 0; i != inputLocationsSize(); ++i){
  auto locations = inputLocations(i);
}
```

test/k4FWCoreTest/src/components/ExampleFunctionalConsumerRuntimeCollections.cpp
m-fila: Yes, I agree, it's complicated to implement and there are already ways to achieve the same result so I would bother with it now. I mentioned it mostly for completeness since they have it in Gaudi.

I could think of something like this: when there multiple arguments like here or more
```cpp
using FloatColl         = std::vector<const podio::UserDataCollection<float>*>;
using ParticleColl      = std::vector<const edm4hep::MCParticleCollection*>;
using SimTrackerHitColl = std::vector<const edm4hep::SimTrackerHitCollection*>;
using TrackerHitColl    = std::vector<const edm4hep::TrackerHit3DCollection*>;
using TrackColl         = std::vector<const edm4hep::TrackCollection*>;

using retType = std::tuple<std::vector<podio::UserDataCollection<float>>, std::vector<edm4hep::MCParticleCollection>,
                           std::vector<edm4hep::MCParticleCollection>, std::vector<edm4hep::SimTrackerHitCollection>,
                           std::vector<edm4hep::TrackerHit3DCollection>, std::vector<edm4hep::TrackCollection>>;

retType operator()(const FloatColl& floatVec, const ParticleColl& particlesVec,
                     const SimTrackerHitColl& simTrackerHitVec, const TrackerHitColl& trackerHitVec,
                     const TrackColl& trackVec) const
```
I think it'd be nice to avoid counting the arguments to write ~~`outputLocations(4)`~~ `outputLocations(5)` or scrolling to the constructor to find the name in `KeyValues`, and instead just write `outputLocations<std::vector<edm4hep::TrackCollection>>()` when I want to get the name for the locations of that "vector of track-collection" output

k4FWCore/include/k4FWCore/Consumer.h
jmcarcell: Changed

k4FWCore/include/k4FWCore/Transformer.h
jmcarcell: Changed

k4FWCore/include/k4FWCore/Transformer.h
jmcarcell: Changed

k4FWCore/include/k4FWCore/Transformer.h
jmcarcell: Changed

k4FWCore/include/k4FWCore/Consumer.h
jmcarcell: It's now there in the message

k4FWCore/include/k4FWCore/Transformer.h
jmcarcell: Ok, I have included an `inputLocationsSize` and `outputLocationsSize` methods.

test/k4FWCoreTest/src/components/ExampleFunctionalConsumerRuntimeCollections.cpp
jmcarcell: I don't have a strong opinion about this, it can be useful. But so far I think we'll have few algorithms with any number of inputs or outputs so the reach will be small.

PR #208, opened 2024-07-05
Add a new test producing histograms and improve comments in tests

jmcarcell:
BEGINRELEASENOTES
- Add a new test producing histograms, making use of the corresponding service in Gaudi
- Improve comments and descriptions in tests

ENDRELEASENOTES


test/k4FWCoreTest/options/CheckOutputFiles.py
```diff
@@ -118,3 +119,13 @@ def check_collections(filename, names):
     "output_k4test_exampledata_functional_mix_iosvc.root",
     mix_collections,
 )
+
+f = ROOT.TFile.Open("functional_producer_hist.root")
+for i in range(2):
+    if (
+        str(f.GetListOfKeys()[i])
```
tmadlener, line 126: Is this guaranteed to be always ordered?

test/k4FWCoreTest/options/ExampleFunctionalProducerHist.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+else:
+    app = ApplicationMgr(
+        TopAlg=[producer1, producer2],
+        EvtSel="NONE",
+        EvtMax=10,
+        ExtSvc=[root_hist_svc] + ([whiteboard] if multithreaded else []),
+        OutputLevel=INFO,
+        HistogramPersistency="ROOT",
+        EventLoop=slimeventloopmgr,
```
tmadlener, line 78: There is no default value for `EventLoop`, that could be used in a similar fashion as the one in `ExtSvc`?

test/k4FWCoreTest/src/components/ExampleFunctionalProducerHist.cpp
```diff
@@ -0,0 +1,49 @@
...
+  // The pair in KeyValues can be changed from python and it corresponds
+  // to the name of the output collection
+  ExampleFunctionalProducerHist(const std::string& name, ISvcLocator* svcLoc)
+      : Producer(name, svcLoc, {}, KeyValues("OutputCollection", {"MCParticles"})) {}
+
+  // This is the function that will be called to produce the data
+  edm4hep::MCParticleCollection operator()() const override {
+    // Not thread-safe!
+    Rndm::Numbers rndu(randSvc(), Rndm::Flat(0, 1));
+    ++m_histograms[rndu()];
```
tmadlener, line 39: Which part here is not thread safe? I suppose it's `randSvc`? Is that only used for seeding here? Can we make this into a thread safe example without too much work?

test/k4FWCoreTest/src/components/ExampleFunctionalProducerHist.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "edm4hep/MCParticleCollection.h"
+
+#include "k4FWCore/Producer.h"
+
+#include "Gaudi/Accumulators/RootHistogram.h"
+#include "GaudiKernel/RndmGenerators.h"
+
+#include <string>
+
+struct ExampleFunctionalProducerHist final : k4FWCore::Producer<edm4hep::MCParticleCollection()> {
```
tmadlener, line 29: Since we only produce an empty collection here, could we not make this a `Consumer` (or something else that doesn't have an output)?

tmadlener (commented):
Very nice. I have a few minor questions inline. I think it would be nice to also have some documentation explaining the main building blocks so that it's a bit more obvious which pieces are related to histograms on the c++ and the python side.

test/k4FWCoreTest/options/ExampleFunctionalProducerHist.py
jmcarcell: I was being lazy, it seems it can only be "EventLoopMgr", anything else won't work for the default.

test/k4FWCoreTest/options/CheckOutputFiles.py
jmcarcell: In the current test I left multithreading disabled not to have too many multithreaded jobs running at same time since `ctest` doesn't know about this but after running it several times I never got the other ordering. I think the histograms may be registered when the class is created and in that case it would always be ordered but I'm not sure.

test/k4FWCoreTest/src/components/ExampleFunctionalProducerHist.cpp
tmadlener: It could take the MCParticles as inputs and then histogram their energy. Or a different datatype for which we already have a functional producer. That would also remove the threadsafety issue with the random numbers(?)

test/k4FWCoreTest/options/CheckOutputFiles.py
tmadlener: The alternative would be to simply make a set of the names in the files and the expected ones and do a set comparison to make sure everything is there to avoid any ordering issues.

test/k4FWCoreTest/src/components/ExampleFunctionalProducerHist.cpp
jmcarcell: I think the engine comes from `randSvc` and since you can do `randSvc()->engine()->setSeeds({1, 2});` any time then I assume it's not going to work. There is something for multithreading here: https://gitlab.cern.ch/gaudi/Gaudi/-/blob/master/GaudiHive/src/HiveNumbers.h#L29 but this is not even being installed and it's only being used in one test in Gaudi.

PR #209, opened 2024-07-15
Add an algorithm for merging collections

jmcarcell:
BEGINRELEASENOTES
- Add an algorithm for merging collections that can merge any number of collections into a new one of the same type. The algorithm supports either creating a subset collection (default) or cloning every object in the collection.
- Add a test using this algorithm to merge a few collections

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/components/CollectionMerger.cpp
+++ b/k4FWCore/components/CollectionMerger.cpp
@@ -27,7 +27,7 @@
 #include <string_view>
 
 struct CollectionMerger final : k4FWCore::Transformer<std::shared_ptr<podio::CollectionBase>(
-                                    const std::map<std::string, const std::shared_ptr<podio::CollectionBase>&>&)> {
+                                    const std::vector<const std::shared_ptr<podio::CollectionBase>*>&)> {
   CollectionMerger(const std::string& name, ISvcLocator* svcLoc)
       : Transformer(name, svcLoc, {KeyValues("InputCollections", {"MCParticles"})},
                     {KeyValues("OutputCollection", {"NewMCParticles"})}) {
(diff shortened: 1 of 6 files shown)


k4FWCore/components/CollectionMerger.cpp
```diff
@@ -0,0 +1,123 @@
...
+  void mergeCollections(const std::shared_ptr<podio::CollectionBase>& source,
+                        std::shared_ptr<podio::CollectionBase>&       ret) const {
+    if (!ret) {
+      ret.reset(new T());
+      if (!m_copy) {
+        ret->setSubsetCollection();
+      }
+    }
+    const auto ptr        = std::dynamic_pointer_cast<T>(ret);
+    const auto sourceColl = std::dynamic_pointer_cast<T>(source);
```
tmadlener, line 110: If we `dynamic_cast` we should also check the results.

tmadlener (commented):
I think we need to very clearly document the potential footguns with `Copy`, mainly that `clone` will only copy data, but will not do anything with the relations. This means that relations in the cloned objects will still point to the where the originals pointed. More importantly: Relations that point *to* the original objects will still point to those and not to the new ones.

jmcarcell:
There are many caveats with this one since you can drop collections that contain the information that you need, with and without copying, so the usage has to be studied a bit more. On relations, I thought about implementing copying of the relations to the new ones but it's not trivial since it has to be done for each collection independently and for now I think I won't implement it unless it's needed. In any case if the usage of this is to use it before an algorithm then to get some collections merged into one and then never use it again it should be fine. If the usage pick collections from a file and make a new file with only the merged one then it won't work as it is.

tmadlener:
Yes, I agree. We should just write these down as long as we still have them in mind. Who knows how many of these we remember in half a year ;)

jmcarcell:
I'll start by merging this one today before there are too many conflicts with other PRs

k4FWCore/components/CollectionMerger.cpp
jmcarcell: These should be static, changed. There should never be any confusion as long as the type one gets with `getTypeName` is the same as the actual type.

PR #210, opened 2024-07-15
Add a filter algorithm

jmcarcell:
that uses `PredicateFilter` from Gaudi. It doesn't read any number of collections, so passing `std::vector<>` will not compile.

BEGINRELEASENOTES
- Add a filter algorithm, using the `PredicateFilter` algorithm from Gaudi.
- Add the minimum handle needed to be able to use the filter from Gaudi and make it work with the rest of the functional algorithms
- Add a test using this filter

ENDRELEASENOTES

@kjvbrt may be interested

### DIFF
--- a/test/k4FWCoreTest/src/components/ExampleFunctionalFilter.cpp
+++ b/test/k4FWCoreTest/src/components/ExampleFunctionalFilter.cpp
@@ -47,7 +47,6 @@ struct ExampleFunctionalFilter final : k4FWCore::FilterPredicate<bool(const edm4
 
   // Thread-safe counter
   mutable Gaudi::Accumulators::Counter<> m_counter{this, "Counter"};
-
 };
 
 DECLARE_COMPONENT(ExampleFunctionalFilter)
--- a/test/k4FWCoreTest/src/components/ExampleFunctionalFilter.cpp
+++ b/test/k4FWCoreTest/src/components/ExampleFunctionalFilter.cpp
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2014-2024 Key4hep-Project.
+ *
+ * This file is part of Key4hep.
+ * See https://key4hep.github.io/key4hep-doc/ for further info.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


test/k4FWCoreTest/src/components/ExampleFunctionalFilter.cpp
```diff
@@ -0,0 +1,52 @@
...
+  ExampleFunctionalFilter(const std::string& name, ISvcLocator* svcLoc)
+      : FilterPredicate(name, svcLoc, KeyValue("InputCollection", {"MCParticles"})) {}
+
+  // This is the function that will be called to check if the event passes
+  // Note that the function has to be const, as well as the collections
+  // we get from the input
+  bool operator()(const edm4hep::MCParticleCollection& input) const override {
+    ++m_counter;
+    // Only pass for half of the events
+    bool condition = m_counter.sum() % 2;
```
tmadlener, line 39: Will this take some lock in the background? (Mainly for my education).

m-fila: After many layers of wrappers there is an atomic operating in`std::memory_order_relaxed`

test/k4FWCoreTest/src/components/ExampleFunctionalFilter.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include "k4FWCore/FilterPredicate.h"
+
+#include <string>
+
+struct ExampleFunctionalFilter final : k4FWCore::FilterPredicate<bool(const edm4hep::MCParticleCollection& input)> {
+  // The pair in KeyValue can be changed from python and it corresponds
+  // to the name of the input collection
+  ExampleFunctionalFilter(const std::string& name, ISvcLocator* svcLoc)
+      : FilterPredicate(name, svcLoc, KeyValue("InputCollection", {"MCParticles"})) {}
```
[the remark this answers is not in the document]
jmcarcell: Some idea I have been thinking about is, after showing here that using the definitions from Gaudi is quite easy, to use for algorithms that don't read or write any number of collections the Gaudi definition, and then the arguments can be strings, but then for the ones with `std::vector` use the definition from `k4FWCore`, all done in the background under the same `k4FWCore::<name>`. This would need some changes to the existing ones but they are not so many.

PR #214, opened 2024-07-19
harmonise two noise tool interfaces

giovannimarchiori:
BEGINRELEASENOTES
- renamed NoiseConstant by NoiseRMS in the noise interfaces (ICaloReadCellNoiseMap, INoiseCaloCellsTool.h and INoiseConstTool.h) to make its physical meaning clearer

ENDRELEASENOTES

Should be merged only when https://github.com/HEP-FCC/k4RecCalorimeter/pull/98 is reviewed/approved and ready to be merged in cascade, which will fix the few occurrences in the code where the method of the interface which changed name is changed (https://github.com/search?q=getNoiseConstantPerCell&type=code)

### DIFF
--- a/k4Interface/include/k4Interface/INoiseConstTool.h
+++ b/k4Interface/include/k4Interface/INoiseConstTool.h
@@ -24,7 +24,7 @@
 
 /** @class INoiseConstTool
  *
- *  Abstract interface to get calorimeter noise per cell tool
+ *  Abstract interface to get calorimeter noise constants (RMS and offset) per cell tool
  *  @author Coralie Neubueser
  *  @date   2018-01
  */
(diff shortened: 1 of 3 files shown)


k4Interface/include/k4Interface/INoiseConstTool.h
```diff
@@ -24,7 +24,7 @@
 
 /** @class INoiseConstTool
  *
- *  Abstract interface to get calorimeter noise per cell tool
+ *  Abstract interface to get calorimeter noise constants (RMS and offset) per cell tool
```
kjvbrt, line 27: Can the naming be also harmonized? I mean to use "noise values" everywhere or "noise constants" :)

PR #215, opened 2024-07-22
Add a metadata service and a test

jmcarcell:
BEGINRELEASENOTES
- Add a metadata service that is automatically added when importing `IOSvc` from `k4FWCore`, if no metadata is added then the service won't do anything. If metadata is added, the metadata service will hold a `podio::Frame` with the metadata that the `Writer` will write. The metadata service only works when there is an output file.
- Add a test showing an example of its usage
- Make changes in the `MetadaHandle` to make it possible to save and read metadata from DataHandle-based algorithms, so that when switching to `IOSvc` metadata for existing DataHandle algorithms will keep working. Add a test using IOSvc with DataHandle algorithms that write and read metadata.

ENDRELEASENOTES


test/k4FWCoreTest/src/components/ExampleFunctionalMetadataConsumer.cpp
```diff
@@ -0,0 +1,83 @@
...
+    m_metadataString = k4FWCore::getParameter<std::string>("MetadataString", this).value_or("");
+    if (m_metadataString != "hello") {
+      error() << "ExampleString is not 'hello'" << endmsg;
+      return StatusCode::FAILURE;
+    }
+    return StatusCode::SUCCESS;
+  }
+
+  void operator()(const edm4hep::MCParticleCollection& input) const override {
+    if (input.size() != m_particleNum) {
```
tmadlener, line 58: Does `k4FWCore::getParameter` also work here?

k4FWCore/include/k4FWCore/IMetadataSvc.h
```diff
@@ -0,0 +1,51 @@
...
+ */
+class IMetadataSvc : virtual public IInterface {
+public:
+  /// InterfaceID
+  DeclareInterfaceID(IMetadataSvc, 1, 0);
+
+  /**
+   * @brief Read the next event from the input file
+   * @return A tuple containing the collections read, the collection names and the frame that owns the collections
+   */
```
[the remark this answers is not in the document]
jmcarcell: Removed

test/k4FWCoreTest/src/components/ExampleFunctionalMetadataProducer.cpp
```diff
@@ -0,0 +1,66 @@
...
+  StatusCode initialize() override {
+    k4FWCore::putParameter("NumberOfParticles", m_particleNum.value(), this);
+    k4FWCore::putParameter("ParticleTime", m_particleTime.value(), this);
+    k4FWCore::putParameter("PDGValues", m_PDGValues.value(), this);
+    k4FWCore::putParameter("MetadataString", m_metadataString.value(), this);
+    return StatusCode::SUCCESS;
+  }
+
+  edm4hep::MCParticleCollection operator()() const override {
+    k4FWCore::putParameter("EventMetadataInt", 5, this);
```
tmadlener, line 43: Where do these parameters go? To the event frame or to the metadata frame? If the latter, I think this should be impossible. At least with the `MetaDataHandle` we went to some length to make this fail (at runtime), because if something is "true metadata" then it should be constant at file level. Hence, there should be no need to make it possible to even set this during event processing. If we can make this call go to the event frame, then it would be OK, I suppose, but it might be a bit confusing to have only one interface for doing both.

The notable exception for setting metadata during event processing is obviously the MarlinWrapper, but I don't think we should make this easily possible here, just because it's needed there.

jmcarcell: This is now forbidden the same way it is done for the `MetaDataHandle` so it should work with the MarlinWrapper.

test/k4FWCoreTest/src/components/ExampleFunctionalMetadataConsumer.cpp
```diff
@@ -0,0 +1,108 @@
...
+#include <vector>
+
+struct ExampleFunctionalMetadataConsumer final : k4FWCore::Consumer<void(const edm4hep::MCParticleCollection& input)> {
+  ExampleFunctionalMetadataConsumer(const std::string& name, ISvcLocator* svcLoc)
+      : Consumer(name, svcLoc, KeyValues("InputCollection", {"MCParticles"})) {}
+
+  StatusCode initialize() override {
+    m_particleNum = k4FWCore::getParameter<int>("NumberOfParticles", this).value_or(0);
+    if (m_particleNum != 3) {
+      error() << "ExampleInt is not 3" << endmsg;
```
andresailer, line 35: One must print out the actual value of the parameter in all these error cases

test/k4FWCoreTest/src/components/ExampleFunctionalMetadataProducer.cpp
jmcarcell: Now changed to prevent from putting parameters in the metadata frame as it was being done for the handle-based algorithms, it's literally the same mechanism so I think it should work for `MarlinWrapper`.

test/k4FWCoreTest/src/components/ExampleFunctionalMetadataConsumer.cpp
jmcarcell: See below

k4FWCore/components/MetadataSvc.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include "MetadataSvc.h"
+
+#include "podio/Frame.h"
+
+#include <GaudiKernel/AnyDataWrapper.h>
+#include <GaudiKernel/IDataProviderSvc.h>
+#include <GaudiKernel/Service.h>
+
+#include "podio/Frame.h"
```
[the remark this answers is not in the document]
jmcarcell: Done

test/k4FWCoreTest/src/components/ExampleFunctionalMetadataConsumer.cpp
jmcarcell: Done

k4FWCore/include/k4FWCore/MetadataUtils.h
```diff
@@ -0,0 +1,49 @@
...
+#include <GaudiKernel/Service.h>
+#include "Gaudi/Algorithm.h"
+
+#include "k4FWCore/IMetadataSvc.h"
+
+// #include "GaudiKernel/CommonMessaging.h"
+
+namespace k4FWCore {
+
+  template <typename T> void putParameter(const std::string& name, const T& value, const Gaudi::Algorithm* alg) {
```
andresailer, line 31: These functions need a docstring, don't they?

test/k4FWCoreTest/options/CheckOutputFiles.py
```diff
@@ -154,3 +158,20 @@ def check_events(filename, number):
...
+        "NumberOfParticles",
+        "ParticleTime",
+        "PDGValues",
+        "MetadataString",
+        "FinalizeMetadataInt",
+    ],
+    [3, 1.5, [1, 2, 3, 4], "hello", 10],
+):
+    if metadata.get_parameter(key) != value:
+        raise RuntimeError(f"Metadata parameter {key} does not match expected value")
```
andresailer, line 177:
```suggestion
        raise RuntimeError(f"Metadata parameter {key} does not match expected value: found {metadata.get_parameter(key)} expected {value}")
```

k4FWCore/include/k4FWCore/MetadataUtils.h
jmcarcell: Docstring added

test/k4FWCoreTest/options/CheckOutputFiles.py
jmcarcell: Done

jmcarcell:
This should now be ready and when using IOSvc it also works fine with DataHandle algorithms. I didn't bother with making it work when using `PodioInput` or `PodioOutput` and functional algorithms since that shouldn't be the case anyway.

andresailer (approved):

PR #217, opened 2024-07-30
Move to new link types to keep compatibility with EDM4hep in the future

tmadlener:
BEGINRELEASENOTES
- Add the legacy `AssociationCollection` headers to keep things building after https://github.com/key4hep/EDM4hep/pull/341 has been merged 

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/components/CollectionMerger.cpp
+++ b/k4FWCore/components/CollectionMerger.cpp
@@ -35,14 +35,6 @@
 
 #include "edm4hep/edm4hep.h"
 
-#include "edm4hep/MCRecoCaloAssociationCollection.h"
-#include "edm4hep/MCRecoCaloParticleAssociationCollection.h"
-#include "edm4hep/MCRecoClusterParticleAssociationCollection.h"
-#include "edm4hep/MCRecoParticleAssociationCollection.h"
-#include "edm4hep/MCRecoTrackParticleAssociationCollection.h"
-#include "edm4hep/MCRecoTrackerAssociationCollection.h"
-#include "edm4hep/RecoParticleVertexAssociationCollection.h"
-
 #include "k4FWCore/Transformer.h"
 
 #include <map>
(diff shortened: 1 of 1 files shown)


tmadlener:
@jmcarcell I would merge this after confirming it doesn't break anything in CI and then trigger the downstream build in https://github.com/key4hep/EDM4hep/pull/341 again

jmcarcell:
Hmm I'm not sure if we should bother with an intermediate step, we are already incompatible with the last release so just switching to the new names shouldn't cause more issues, I guess you just want to see how it affects other repos? But then with the workflow is going to take a while :smiley:

tmadlener:
I can also just move to the new types directly, and check locally. That would then tie this PR strictly to the EDM4hep one.

jmcarcell:
Yeah, I think in general since we are not compatible anyway with an older version we can move directly for this and other repos. More work upfront but it will be done.

k4FWCore/components/CollectionMerger.cpp
```diff
@@ -69,20 +69,20 @@ struct CollectionMerger final : k4FWCore::Transformer<std::shared_ptr<podio::Col
...
-        &CollectionMerger::mergeCollections<edm4hep::MCRecoTrackerAssociationCollection>;
-    m_map["edm4hep::MCRecoCaloParticleAssociationCollection"] =
-        &CollectionMerger::mergeCollections<edm4hep::MCRecoCaloParticleAssociationCollection>;
-    m_map["edm4hep::MCRecoClusterParticleAssociationCollection"] =
-        &CollectionMerger::mergeCollections<edm4hep::MCRecoClusterParticleAssociationCollection>;
-    m_map["edm4hep::MCRecoTrackParticleAssociationCollection"] =
-        &CollectionMerger::mergeCollections<edm4hep::MCRecoTrackParticleAssociationCollection>;
-    m_map["edm4hep::RecoParticleVertexAssociationCollection"] =
-        &CollectionMerger::mergeCollections<edm4hep::RecoParticleVertexAssociationCollection>;
+    m_map["edm4hep::RecoMCParticleLinkCollection"] =
```
jmcarcell, line 72: Here one could think that duplicating the lines to also have the old names in the map (mapped to merging the new ones) could be useful but files with the old names can have been produced with the latest release and are incompatible or with an older nightly which will also be incompatible.

jmcarcell (approved):

tmadlener:
Things are building locally with this and key4hep/EDM4hep#341 for me. So from my side we can go ahead and merge them both.

k4FWCore/components/CollectionMerger.cpp
tmadlener: In principle the new typenames should be enough because they come from the collection in any case. So even if an older version / name was on disk, by the time it reaches this place, we will only ever have to deal with the actual / new typename.

PR #222, opened 2024-08-11
Fix warnings when warnings are enabled

jmcarcell:
I was compiling k4Reco and getting warnings about files in k4FWCore and it turns out that warnings are not enabled in k4FWCore but are in k4Reco. I propose a general solution for all the Key4hep repositories in https://github.com/key4hep/key4hep-dev-utils/pull/5 so feel free to have a look and comment.

BEGINRELEASENOTES
- Fix warnings when warnings are enabled. Warnings will be enabled later.

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/include/k4FWCore/Consumer.h
+++ b/k4FWCore/include/k4FWCore/Consumer.h
@@ -98,7 +98,7 @@ namespace k4FWCore {
        * @param i  The index of the input
        * @return   A range of the input locations
        */
-      const auto inputLocations(int i) const {
+      auto inputLocations(unsigned i) const {
         if (i >= sizeof...(In)) {
           throw std::out_of_range("Called inputLocations with an index out of range, index: " + std::to_string(i) +
                                   ", number of inputs: " + std::to_string(sizeof...(In)));
@@ -110,7 +110,7 @@ namespace k4FWCore {
        * @param name  The name of the input
        * @return      A range of the input locations
        */
-      const auto inputLocations(std::string_view name) const {
+      auto inputLocations(std::string_view name) const {
         auto it = std::ranges::find_if(m_inputLocations, [&name](const auto& prop) { return prop.name() == name; });
         if (it == m_inputLocations.end()) {
           throw std::runtime_error("Called inputLocations with an unknown name");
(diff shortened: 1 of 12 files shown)


k4FWCore/include/k4FWCore/Transformer.h
```diff
@@ -131,7 +131,7 @@ namespace k4FWCore {
        * @param i  The index of the input
        * @return   A range of the input locations
        */
-      auto inputLocations(int i) const {
+      auto inputLocations(unsigned i) const {
```
m-fila, line 134:
```suggestion
      auto inputLocations(size_t i) const {
```

k4FWCore/include/k4FWCore/Consumer.h
```diff
@@ -98,7 +98,7 @@ namespace k4FWCore {
        * @param i  The index of the input
        * @return   A range of the input locations
        */
-      const auto inputLocations(int i) const {
+      auto inputLocations(unsigned i) const {
```
m-fila, line 101: how about using size_t since this is later compared with sizeof... (gives size_t) and used as index?
```suggestion
      auto inputLocations(size_t i) const {
```

test/k4FWCoreTest/src/components/ExampleFunctionalConsumerRuntimeCollectionsMultiple.cpp
```diff
@@ -42,6 +42,8 @@ struct ExampleFunctionalConsumerRuntimeCollectionsMultiple final
   void operator()(const std::vector<const edm4hep::MCParticleCollection*>& particleVec,
                   const std::vector<const edm4hep::TrackCollection*>&      trackVec,
                   const edm4hep::SimTrackerHitCollection&                  simTrackerHits) const override {
+    // Silence warning about unused variable
+    (void)simTrackerHits;
```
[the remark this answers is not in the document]
jmcarcell: Since these examples are copied and pasted often I did it on purpose but removing the variable is also fine

m-fila:
Looks like the `size_t` suggestions were applied and then disappeared during rebase (in case that was by accident)

jmcarcell:
Should be fixed now, I must have rebased without pulling

PR #224, opened 2024-08-20
Fix data race in EventHeaderCreator 

m-fila:
BEGINRELEASENOTES
- Fixed data race in `EventHeaderCreator` and make it ready for GaudiHive 

ENDRELEASENOTES

A `static` local variable was not thread safe and resulting in a data race when used in parallel environment of GaudiHive (skipping or duplicating event numbers). 

I considered two ways to fix it:
- use `EventContex` to get the event number as seen by scheduler. Currently supplying the `EventContex` as input to `k4FWCore::Transformer` is not allowed so the algorithm will stay in non-functional form. If there is any filtering done by some algorithms earlier in the sequence then this will be not reflected in the event number as it's determined as the scheduler starts this event. If two events are loaded from input the ordering of event numbers given to them will be the same as ordering of loading.
- use atomic member variable as counter. If there is any filtering done by some algorithms earlier in the sequence then this will be reflected in the event numbers as the event number is determined only when this algorithm is executed. If two events are loaded from input the ordering of event numbers given to them doesn't have to be the same as the order of loading - it will be the same as order of execution which might be different

In both cases the event numbers in the output will be not sorted.

The version with with `EventContext` is used currently but can be changed if needed


k4FWCore/components/EventHeaderCreator.cpp
```diff
@@ -26,25 +26,13 @@ EventHeaderCreator::EventHeaderCreator(const std::string& name, ISvcLocator* svc
...
-StatusCode EventHeaderCreator::initialize() {
-  if (Gaudi::Algorithm::initialize().isFailure())
-    return StatusCode::FAILURE;
-  return StatusCode::SUCCESS;
-}
-
-StatusCode EventHeaderCreator::execute(const EventContext&) const {
-  static int eventNumber = 0;
+StatusCode EventHeaderCreator::execute(const EventContext& ctx) const {
+  auto eventNumber = ctx.evt();
```
andresailer, line 30: Who sets the event number in this case? And to what values under what circumstances? Is this deterministic and reproducible?

m-fila: It is ordinal event number created by the `EventProcessor` (event loop).  Both `HiveSlimEventLoopMgr` and `EventLoopMgr` keep track of the number of events run so far, update the number in `nextEvent` and put it in `EventContext`

Being deterministic and reproducible depends on the scheduler. 
- With event context solution that would mean there is no another `Reader` alg executed (from different event processed in parallel) between `EventProcessor::nextEvent()` and `Reader` algorithm for a given event.
- With  the atomic counter solution that would mean there is no `EventHeaderCreator` alg executed between `Reader` alg and `EventHeaderCreator` alg for a given event

andresailer: With "there is no", do you mean "there must be no"? Can this be guaranteed?

m-fila: Sorry, I meant there should be no overlap between those operations in order to have no mismatch between ordinal number of event in an input file and number used in header.

For example if there are just three events in an input file and the `EventHeaderCreator` is used to create the headers for them , then if there is an overlap the between those operations then the event numbers in headers could be any permutation of (0,1,2) with no reproducibility (but at least there will be no duplicates as in the current `EventHeaderCreator`)

 I don't think these can be guaranteed with `AvalancheSchedulerSvc` (beside limiting the thread pool and concurrency slots)

test/k4FWCoreTest/src/components/ExampleEventHeaderCheck.cpp
```diff
@@ -0,0 +1,68 @@
+/*
```
jmcarcell, line 1: It's a bit overkill to create another algorithm just to check that the event numbers are a permutation of some numbers, no? We could test this with the python script to test outputs.

m-fila: Yeah I went overboard with this. Moved testing to `checkOutputFiles.py`

k4FWCore/components/EventHeaderCreator.cpp
m-fila: There was a discussion on the key4hep meeting a week ago that it should be acceptable to not have deterministic event ids when running in a multithreaded environment

PR #228, opened 2024-08-23
fix virtual specifiers, remove redundant methods, unclutter

m-fila:
BEGINRELEASENOTES
- fixed virtual specifiers, removed redundant methods, uncluttered code

ENDRELEASENOTES

This is an attempt to slightly reduce the amount of dated and obsolete code:
 - missing or redundant virtual, override, final specifiers
 -  redundant declarations of trivial virtual destruction when base method already has virtual destructor,
 - removed redundant empty destructors
 - use `empty()` methods instead of comparison for `vector` and `string`
 - use make smart pointer instead of `reset(new `
 - use auto instead of repeating a type or type long iterator types

### DIFF
--- a/k4FWCore/components/CollectionMerger.cpp
+++ b/k4FWCore/components/CollectionMerger.cpp
@@ -121,7 +121,7 @@ private:
   void mergeCollections(const std::shared_ptr<podio::CollectionBase>& source,
                         std::shared_ptr<podio::CollectionBase>&       ret) const {
     if (!ret) {
-      ret.reset(new T());
+      ret = std::make_shared<T>();
       if (!m_copy) {
         ret->setSubsetCollection();
       }
--- a/k4FWCore/components/EventHeaderCreator.cpp
+++ b/k4FWCore/components/EventHeaderCreator.cpp
@@ -26,12 +26,6 @@ EventHeaderCreator::EventHeaderCreator(const std::string& name, ISvcLocator* svc
                   "Name of the EventHeaderCollection that will be stored in the output root file.");
 }
 
-StatusCode EventHeaderCreator::initialize() {
-  if (Gaudi::Algorithm::initialize().isFailure())
-    return StatusCode::FAILURE;
-  return StatusCode::SUCCESS;
-}
-
 StatusCode EventHeaderCreator::execute(const EventContext&) const {
   static int eventNumber = 0;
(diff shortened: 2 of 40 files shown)


m-fila:
Thank you
> Should we explicitly `default` the empty destructors that we would otherwise remove here? It's not necessary from a c++ point of view, but it would make it very clear that nothing special is happening in this class destructor from just looking at the code.

I'd prefer to use implicit destructors when possible in the spirit of [rule 0](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-zero) (having `= default` reads like there was a need to be explicit because the class does something not trivial with memory). But if there is a strong preference to have explicit destructors in the codebase then I can bring them back no problem

>We seem to have an arbitrary mix of `override` and `final` in our algorithms and classes. I think all algorithms could probably be `final` (especially the test and example ones). It's not very likely that something will inherit from them. For all the services a similar argument holds probably, so we could make almost everything `final`(?). This doesn't matter too much with functional algorithms, so I am not sure whether it's worth the effort in the end.

I was wondering if there is some conventions/deeper reason for the mix. Is there really a reason to not allow further overriding (`final` doesn't have performance benefits and to me most of )? In that case maybe the classes should be `final`?

andresailer (approved):

tmadlener:
> I'd prefer to use implicit destructors when possible in the spirit of [rule 0](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-zero) (having = default reads like there was a need to be explicit because the class does something not trivial with memory). But if there is a strong preference to have explicit destructors in the codebase then I can bring them back no problem

Good point. I haven't really checked the existence of the constructors. So if we can go with *rule of 0* let's go with that.

> I was wondering if there is some conventions/deeper reason for the mix. Is there really a reason to not allow further overriding (`final` doesn't have performance benefits and to me most of )? In that case maybe the classes should be final?

Maybe it's time to come up with a convention and write it down into some general guidelines document.

jmcarcell:
Anything left here or can we merge this @m-fila?

m-fila:
It's good to go. I think we could propose a convention for `final` later as no-one seems to have now a very strong opinion

jmcarcell:
I agree with algorithms being final, for services we should check if no one has derived another service from any of them.

PR #230, opened 2024-09-03
cmake: Overhaul downstream config to properly export version and dependencies

### LINKED ISSUE #166: Missing version information in CMake configuration
There is currently no version information set in the cmake config, something like:

```cmake
# project version
SET( ${PROJECT_NAME}_VERSION_MAJOR X )
SET( ${PROJECT_NAME}_VERSION_MINOR Y )
SET( ${PROJECT_NAME}_VERSION_PATCH Z )

SET( ${PROJECT_NAME}_VERSION  "${${PROJECT_NAME}_VERSION_MAJOR}.${${PROJECT_NAME}_VERSION_MINOR}.${${PROJECT_NAME}_VERSION_PATCH}" )
```

This makes usage of the tagging script impossible at the moment, and also just makes it hard to use this package downstream in a versioned way.

tmadlener:
BEGINRELEASENOTES
- Configure the `k4FWCoreConfig.cmake` to put a bit more information in there
  - Make sure all dependencies are also found for dependent packages
  - Make sure dependency versions are discovered again consistently
  - Make sure to export the current k4FWCore version for downstream consumers

ENDRELEASENOTES

There are a few things that we could also address while we are here
- Currently we don't export any version information at all. Since we are now in iteration 20 of the `1.0` pre-release cycle, maybe we should start to do so (and then also propagate that information to downstream consumers)
- ~Is podio 0.16.3 still something that we can use? With the addition of the `MetadataSvc` (#215 and #223), I think we need at least podio 1.0 in any case.~

Fixes #166


cmake/k4FWCoreCreateConfig.cmake
```diff
@@ -0,0 +1,6 @@
+include(CMakePackageConfigHelpers)
+
+configure_package_config_file(${PROJECT_SOURCE_DIR}/cmake/k4FWCoreConfig.cmake.in
+                              ${PROJECT_BINARY_DIR}/k4FWCoreConfig.cmake
+                              INSTALL_DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/k4FWCore
+                              PATH_VARS CMAKE_INSTALL_INCLUDEDIR CMAKE_INSTALL_LIBDIR)
```
jmcarcell, line 6: I would put these in the main `CMakeLists.txt`, it's only a few more lines.

tmadlener: It will become a few more once we start to also export version information properly. Since we have a usable tag already, should I add that as well?

jmcarcell: Yeah, why not

CMakeLists.txt
```diff
@@ -59,5 +66,6 @@ install(EXPORT ${PROJECT_NAME}Targets
   DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}/"
   )
 
-gaudi_install(CMAKE cmake/${PROJECT_NAME}Config.cmake)
+include(cmake/k4FWCoreCreateConfig.cmake)
 
+gaudi_install(CMAKE ${PROJECT_BINARY_DIR}/k4FWCoreConfig.cmake)
```
jmcarcell, line 71:
```suggestion
install(FILES ${PROJECT_BINARY_DIR}/k4FWCoreConfig.cmake DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/k4FWCore)
```
It's a single cmake file, `gaudi_install` doesn't do anything special for us and this is clearer (and we can check that the paths here and given to `configure_package_config_file` match).

cmake/k4FWCoreCreateConfig.cmake
tmadlener: The current 1.1.2 tag is now used.

CMakeLists.txt
tmadlener: Done.

test/downstream-project-cmake-test/CMakeLists.txt
```diff
@@ -23,5 +23,13 @@ project(DownstreamProjectUsingk4FWCore)
 
 find_package(k4FWCore)
 
+if (NOT DEFINED ${k4FWCore_VERSION})
+  message(FATAL_ERROR "CAnnot find k4FWCore version information")
```
m-fila, line 27:
```suggestion
  message(FATAL_ERROR "Cannot find k4FWCore version information")
```
Typo

tmadlener: Thanks :)

jmcarcell (approved):

PR #231, opened 2024-09-03
Fix accessing metadata with `MetaDataHandle` when `MetadataSvc` is present

m-fila:
BEGINRELEASENOTES
- Fixed missing checks for accessing metadata with `MetaDataHandle` when used with `IOSvc`

ENDRELEASENOTES

Fixes for `MetaDataHandle` interoperability with `PodioDataSvc` and new `MetadataSvc`/`IOSvc` introduced in #215:
- `MetaDataHandle::get_optional` wasn't checking whether `PodioDataSvc` or `IOSvc` is used
- `MetaDataHandle::get` didn't check validity of optional when used with `IOSvc`

And some additional cleanup:
- removed constructor with missing definition (unused otherwise that would be an error)
- removed workaround for checking service existence. Gaudi's `service` by default initializes a service if it's missing, hence the check was always true

---

This is PR is orthogonal and doesn't depend on #223

### DIFF
--- a/test/k4FWCoreTest/src/components/k4FWCoreTest_cellID_reader.cpp
+++ b/test/k4FWCoreTest/src/components/k4FWCoreTest_cellID_reader.cpp
@@ -42,12 +42,12 @@ StatusCode k4FWCoreTest_cellID_reader::execute(const EventContext&) const {
   auto optCellIDstr = m_cellIDHandle.get_optional();
 
   if (!optCellIDstr.has_value()) {
-    error() << "ERROR cellID is empty but was expected to hold a value" << endmsg;
+    error() << "ERROR: cellID is empty but was expected to hold a value" << endmsg;
     return StatusCode::FAILURE;
   }
 
   if (optCellIDstr.value() != cellIDstr) {
-    error() << "ERROR metadata accessed with by optional and value differs" << endmsg;
+    error() << "ERROR: metadata accessed with by optional and value differs" << endmsg;
     return StatusCode::FAILURE;
   }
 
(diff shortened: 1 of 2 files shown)


test/k4FWCoreTest/src/components/k4FWCoreTest_cellID_reader.cpp
```diff
@@ -39,5 +39,17 @@ StatusCode k4FWCoreTest_cellID_reader::execute(const EventContext&) const {
     return StatusCode::FAILURE;
   }
 
+  auto optCellIDstr = m_cellIDHandle.get_optional();
+
+  if (!optCellIDstr.has_value()) {
+    error() << "ERROR cellID is empty but was expected to hold a value" << endmsg;
```
jmcarcell, line 45:
```suggestion
    error() << "ERROR: cellID is empty but was expected to hold a value" << endmsg;
```

test/k4FWCoreTest/src/components/k4FWCoreTest_cellID_reader.cpp
```diff
@@ -39,5 +39,17 @@ StatusCode k4FWCoreTest_cellID_reader::execute(const EventContext&) const {
...
+    return StatusCode::FAILURE;
+  }
+
+  if (optCellIDstr.value() != cellIDstr) {
+    error() << "ERROR metadata accessed with by optional and value differs" << endmsg;
```
jmcarcell, line 50:
```suggestion
    error() << "ERROR: metadata accessed with by optional and value differs" << endmsg;
```

jmcarcell (approved):

PR #233, opened 2024-09-06
Use podio::Reader and podio::Writer with IOSvc

jmcarcell:
BEGINRELEASENOTES
- Use podio::Reader and podio::Writer with IOSvc, so that it is easy to write TTrees or RNTuples by changing the `IOType` parameter given to `IOSvc` in the steering file.

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/components/IIOSvc.h
+++ b/k4FWCore/components/IIOSvc.h
@@ -22,7 +22,7 @@
 #include "GaudiKernel/IInterface.h"
 
 #include "podio/CollectionBase.h"
-#include "podio/ROOTWriter.h"
+#include "podio/Writer.h"
 
 #include <memory>
 #include <vector>
@@ -48,10 +48,10 @@ public:
                                                     next()                     = 0;
   virtual std::shared_ptr<std::vector<std::string>> getCollectionNames() const = 0;
 
-  virtual std::shared_ptr<podio::ROOTWriter> getWriter()                                         = 0;
-  virtual void                               deleteWriter()                                      = 0;
-  virtual void                               deleteReader()                                      = 0;
-  virtual bool                               checkIfWriteCollection(const std::string& collName) = 0;
+  virtual std::shared_ptr<podio::Writer> getWriter()                                         = 0;
+  virtual void                           deleteWriter()                                      = 0;
+  virtual void                           deleteReader()                                      = 0;
+  virtual bool                           checkIfWriteCollection(const std::string& collName) = 0;
 };
 
 #endif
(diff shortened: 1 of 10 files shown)


k4FWCore/components/IOSvc.cpp
```diff
@@ -42,6 +42,9 @@ StatusCode IOSvc::initialize() {
   if (!m_importedFromk4FWCore) {
     error() << "Use 'from k4FWCore import IOSvc' instead of 'from Configurables import IOSvc' to access the service"
             << endmsg;
+  }
+  if (m_outputType != "default" && m_outputType != "ROOT" && m_outputType != "RNTuple") {
+    error() << "Unknown input type: " << m_outputType << ", expected ROOT, RNTuple or default" << endmsg;
```
m-fila, line 47: `podio::Writer` allows also SIO. Is this intentional to exclude it?

m-fila (commented):
Is there a reason to dynamically allocate the writer and reader?
- They already are type erased
- `getWriter()` always return a valid object and is always assumed to do so
- I think the wrtier's `finish` method could be called directly instead triggering with pointer reset, no?

k4FWCore/components/IOSvc.cpp
jmcarcell: The demand for writing EDM4hep files using SIO is quite low. If someone wants it we can add it.

tmadlener: Could be nice for some testing at some point, but I agree we can easily toggle this once it becomes necessary.

jmcarcell:
Now the `Reader` and `Writer` are stored in an `std::optional` and there isn't any dynamic allocation.
```
I think the wrtier's finish method could be called directly instead triggering with pointer reset, no?
```
We could but then the writer can't be used anymore so we may as well delete it, which prevents further usage of the writer possibly by mistake.

k4FWCore/components/IIOSvc.h
```diff
@@ -48,10 +43,10 @@ class IIOSvc : virtual public IInterface {
                                                     next()                     = 0;
   virtual std::shared_ptr<std::vector<std::string>> getCollectionNames() const = 0;
 
-  virtual std::shared_ptr<podio::ROOTWriter> getWriter()                                         = 0;
-  virtual void                               deleteWriter()                                      = 0;
-  virtual void                               deleteReader()                                      = 0;
-  virtual bool                               checkIfWriteCollection(const std::string& collName) = 0;
+  virtual podio::Writer& getWriter()                                         = 0;
+  virtual void           deleteWriter()                                      = 0;
+  virtual void           deleteReader()                                      = 0;
```
tmadlener, line 48: Should these be named differently now that they are no longer pointer backed? Something like `resetWriter|resetReader`?

jmcarcell: I think it doesn't matter but the reason why they are needed is because services in Gaudi are not deleted so making the connection would be easier and technically they are being deleted.

PR #234, opened 2024-09-06
Add the possibility of starting from the Nth event instead of the first one

jmcarcell:
BEGINRELEASENOTES
- Add the possibility of starting from the Nth event instead of the first one with the argument `FirstEventEntry` to `IOSvc`.
-  Add a test that ignores some events.

ENDRELEASENOTES

Someone asked for this and it was possible before with `PodioInput`.


tmadlener:
It can be quite useful for debugging.

andresailer:
Could we add the relevant argument to the release notes entry?

test/k4FWCoreTest/options/ExampleFunctionalReadNthEvent.py
```diff
@@ -0,0 +1,41 @@
...
+
+from Gaudi.Configuration import INFO
+from Configurables import ExampleFunctionalTransformer
+from Configurables import EventDataSvc
+from k4FWCore import ApplicationMgr, IOSvc
+
+svc = IOSvc("IOSvc")
+svc.Input = "functional_producer.root"
+svc.Output = "functional_nth_event.root"
+svc.FirstEventEntry = 5
```
m-fila, line 31: Could it be changed to some other number so it's easier to reason? The input has 10 entries, the first entry is 5, the output has 5 entries. 6-4 or any other split would read better

jmcarcell: Done

m-fila (approved):

PR #236, opened 2024-09-16
Add an argument for add_test_with_env not to have to repeat the test names

jmcarcell:
BEGINRELEASENOTES
- Add an argument for add_test_with_env not to have to repeat the test names
- Also add some documentation for the function since it's getting complicated

ENDRELEASENOTES

No more endless horizontal scrolling


m-fila:
Would it make sense to move `CheckOutputFiles.py` from `options/ as it's not a steering file? To  `scripts/` for instance?

jmcarcell:
Makes sense but we install the whole scripts folder, which I don't think it's a bad idea, and then this file would also be installed. I think it's fine as it is since it's only a folder for tests.

m-fila:
Sorry I meant to `test/k4FWCoreTest/scripts`, that isn't installed, right?

jmcarcell:
Yes, moved.

PR #237, opened 2024-09-18
Assign to a const char* when using std::getenv

jmcarcell:
BEGINRELEASENOTES
- Assign to a const char* when using std::getenv

ENDRELEASENOTES

Discovered after https://github.com/AIDASoft/podio/pull/676, it turns out that the one in Writer.cpp was not being assigned to `const char*` and it failed (it doesn't fail anymore after https://github.com/AIDASoft/podio/pull/680). Because `const auto` translates to `char* const` while `const auto*` translates to `const char*`. Use `*`!

### DIFF
--- a/k4FWCore/components/PodioOutput.cpp
+++ b/k4FWCore/components/PodioOutput.cpp
@@ -140,7 +140,7 @@ StatusCode PodioOutput::finalize() {
   // Collect all the metadata
   podio::Frame config_metadata_frame{};
   config_metadata_frame.putParameter("gaudiConfigOptions", config_data);
-  if (const auto* env_key4hep_stack = std::getenv("KEY4HEP_STACK")) {
+  if (const char* env_key4hep_stack = std::getenv("KEY4HEP_STACK")) {
     std::string s_env_key4hep_stack = env_key4hep_stack;
     config_metadata_frame.putParameter("key4hepstack", s_env_key4hep_stack);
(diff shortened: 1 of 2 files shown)


andresailer:
None of the cases you describe correspond to what was there before?
`const char*` is supposed to be what is good, but it is changed to `const auto*`
and the other case is just `auto` which isn't in the list of things you describe what it evaluates to?

Why not be explicit and use `char` instead of `auto`? But at least align the release notes with what you do, which is assigning to `const auto*`?

jmcarcell:
> What does "fail" mean in this case? It stored a wrong value? Or it didn't store anything because the whole if condition was evaluated as `false`?

Fail means failed to compile because of the `static_assert` in https://github.com/AIDASoft/podio/pull/676.



> None of the cases you describe correspond to what was there before? `const char*` is supposed to be what is good, but it is changed to `const auto*` and the other case is just `auto` which isn't in the list of things you describe what it evaluates to?
> 
> Why not be explicit and use `char` instead of `auto`? But at least align the release notes with what you do, which is assigning to `const auto*`?

I thought it was `const auto` and then the point I was making was that only `const auto*` will work with https://github.com/AIDASoft/podio/pull/676. Anyway now it's `const char*`.

tmadlener (approved):

andresailer (approved):

Issue #241, opened 2024-09-26
IOSvc.Input property can't be set from CLI


m-fila:
I cannot pass to k4run the input files to process via --IOSvc.input blah.root. On the other hand IOSvc.output foo.root works like a charm. The difference between the two is that output is a string property while input is a vector<string> property, so the problem seems to be in IOSvc parsing of vector<..>. Is there a special syntax to be used in the CLI that I am unaware of or is it a parsing problem in IOSvc?

_Originally posted by @giovannimarchiori in https://github.com/key4hep/k4FWCore/issues/238#issuecomment-2376461994_

m-fila:
The `IOSvc.Input` is not listed in the acceptable arguments (unless it's assigned in a steering file):
```
  --IOSvc.OutputLevel [IOSVC.OUTPUTLEVEL], --OutputLevel.IOSvc [IOSVC.OUTPUTLEVEL]
                        output level [Service]
  --IOSvc.output [IOSVC.OUTPUT], --output.IOSvc [IOSVC.OUTPUT]
                        List of files to write output to [IOSvc]
  --IOSvc.Output [IOSVC.OUTPUT], --Output.IOSvc [IOSVC.OUTPUT]
                        List of files to write output to [IOSvc]
  --IOSvc.outputCommands IOSVC.OUTPUTCOMMANDS [IOSVC.OUTPUTCOMMANDS ...], --outputCommands.IOSvc IOSVC.OUTPUTCOMMANDS [IOSVC.OUTPUTCOMMANDS ...]
                        A set of commands to declare which collections to keep or drop. [IOSvc]
  --IOSvc.IOType [IOSVC.IOTYPE], --IOType.IOSvc [IOSVC.IOTYPE]
                        Type of input file (ROOT, RNTuple) [IOSvc]
  --IOSvc.ImportedFromk4FWCore [IOSVC.IMPORTEDFROMK4FWCORE], --ImportedFromk4FWCore.IOSvc [IOSVC.IMPORTEDFROMK4FWCORE]
                        This is set to true when IOSvc is imported from k4FWCore instead of Configurables in python [IOSvc]
  --IOSvc.FirstEventEntry [IOSVC.FIRSTEVENTENTRY], --FirstEventEntry.IOSvc [IOSVC.FIRSTEVENTENTRY]
                        First event entry to read [IOSvc]
```

[...]

PR #242, opened 2024-09-27
Modernize and clean up k4run

jmcarcell:
Fixes https://github.com/key4hep/k4FWCore/issues/241 and does a much needed clean up of the script.

BEGINRELEASENOTES
- Add a docstring for the function that adds the arguments
- Remove dead code, obvious comments and global variables
- Use f-strings
- Fix arguments when a property is a std::vector. Related to this fix now `k4run` will also display properties for which the default value is an empty list, previously these were not displayed.
- Improve formatting of the message with all the values of all the properties

ENDRELEASENOTES

Now the messages at startup with all the options are cleaner. Before:

```
[k4run] Option name: Transformer.OutputLevel Transformer.OutputLevel 0
[k4run] Option name: Transformer.Asynchronous Transformer.Asynchronous False
[k4run] Option name: Transformer.InputCollection Transformer.InputCollection ['MCParticles']
[k4run] Option name: Transformer.OutputCollection Transformer.OutputCollection ['NewMCParticles']
[k4run] Option name: Transformer.Offset Transformer.Offset 10
```

after:

```
[k4run] Option name: Transformer.OutputLevel 0
[k4run] Option name: Transformer.Asynchronous False
[k4run] Option name: Transformer.InputCollection ['MCParticles']
[k4run] Option name: Transformer.OutputCollection ['NewMCParticles']
[k4run] Option name: Transformer.Offset 10
```

### DIFF
--- a/k4FWCore/scripts/k4run
+++ b/k4FWCore/scripts/k4run
@@ -70,20 +70,19 @@ def add_arguments(parser, app_mgr):
     for conf in frozenset(app_mgr.allConfigurables.values()):
         if conf.name() in ["ApplicationMgr", "ToolSvc", "k4FWCore__Sequencer", "k4FWCore__Algs"]:
             continue
-        props = conf.getPropertiesWithDescription()  # property values -> property description
-        for prop in props:
+        for prop_name, prop_value in conf.getPropertiesWithDescription().items():
             if (
-                prop in FILTER_GAUDI_PROPS
-                or "Audit" in prop
-                or hasattr(props[prop][0], "__slots__")
+                prop_name in FILTER_GAUDI_PROPS
+                or "Audit" in prop_name
+                or hasattr(prop_value[0], "__slots__")
             ):
                 continue
-            value = props[prop][0]
+            value = prop_value[0]
 
             # if it is set to "no value" it hasn't been touched in the options file
             if value == conf.propertyNoValue:
-                value = conf.getDefaultProperty(prop)
-            proptype = type(props[prop][0])
+                value = conf.getDefaultProperty(prop_name)
+            proptype = type(prop_value[0])
             args = "?"
             if proptype is list:
                 if value:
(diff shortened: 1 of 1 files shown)


k4FWCore/scripts/k4run
```diff
@@ -45,62 +45,65 @@ FILTER_GAUDI_PROPS = [
...
-seen_files = set()
-option_db = {}
+def add_arguments(parser, app_mgr):
+    """
+        Add arguments to the parser for all properties of all configurables in the application manager
 
+        :param parser: the parser to add the arguments to
+        :param app_mgr: the application manager to get the properties from
+
+        :return: a dictionary mapping the argument name to the configurable it belongs to
```
tmadlener, line 56: Point for discussion here: Which docstring style do we want to use in Key4hep? I don't think we have consistently decided this (or even discussed this). In podio I opted for  [google's style for docstrings](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings)

k4FWCore/scripts/k4run
```diff
@@ -45,62 +45,65 @@ FILTER_GAUDI_PROPS = [
...
-        # dict propertyname: (propertyvalue, propertydescription)
-        props = conf.getPropertiesWithDescription()
+        props = conf.getPropertiesWithDescription()  # property values -> property description
         for prop in props:
             if (
                 prop in FILTER_GAUDI_PROPS
                 or "Audit" in prop
                 or hasattr(props[prop][0], "__slots__")
             ):
                 continue
```
tmadlener, line 80: Assuming `props` is a dictionary, this whole loop could be simplified via
```suggestion
        for prop_name, prop_values in props.items():
            if (
                prop_name in FILTER_GAUDI_PROPS
                or "Audit" in prop_name
                or hasattr(prop_values[0], "__slots__")
            ):
                continue
```
and then similar for the rest of the loop. (github's UI doesn't allow to easily make suggestions over areas with mixed diffs).

jmcarcell: Done

k4FWCore/scripts/k4run
jmcarcell: I don't mind. We can discuss this where some default configuration would be placed (https://github.com/key4hep/key4hep-dev-utils).

jmcarcell:
If there aren't any other comments or complains I'll merge this soon since I want this fix to be included in the next tag that is coming soon.

PR #243, opened 2024-10-01
Fix overwritting the input file for IOSvc from the command line

jmcarcell:
Originally posted in https://github.com/key4hep/k4FWCore/issues/241#issuecomment-2384294263

BEGINRELEASENOTES
- Fix overwritting the input file for IOSvc from the command line. The k4FWCore wrapper of the ApplicationMgr checks the number of events (needed for multithreading) and if the file exists but this was happening right after the parsing of the file, without taking into account the input file can be overwritten from the command line.
- Add a test where the input file is being overwritten from the command line

ENDRELEASENOTES


m-fila:
For completeness another test could be added checking the case whe `IOSvc.Input` is not assigned in the steering file but set from cli

giovannimarchiori:
Thanks, I tested it locally and it works (party!)

jmcarcell:
I think this should be ready; I have added another check for reading from multiple files. I'm not aware of any other bugs and passing through the command line overwrites the choice in the steering file (if any) correctly

PR #244, opened 2024-10-01
Update README and Use Key4hep instead of Key4HEP

jmcarcell:
BEGINRELEASENOTES
- Update README: remove note about GaudiAlg being outdated since it's not there anymore and Use Key4hep instead of Key4HEP

ENDRELEASENOTES

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,9 @@ Algorithm to write data to an output file on disk.
 ## k4run
 ```
 $ k4run --help
-usage: k4run [--dry-run] [-v] [-n NUM_EVENTS] [-l] [--gdb] [-h] [config_files ...]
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -77,6 +77,8 @@ types of algorithms, depending on your use case:
 - The `Transformer` is the more general one (both the `Consumer` and the
   `Producer` are a particular case of this one) and takes both inputs and
   outputs
+- The `Filter`, for removing events from the events that are going to be
```
m-fila, line 80:
```suggestion
- The `FilterPredicate`, for removing events from the events that are going to be
```

m-fila: I think this explanation is a bit misleading since, the even will be marked to be filtered out only in the current branch of the workflow (and whether it's dropped there depends on the properties of the Sequencer/DecisionHub to which it belongs)

How about something like this:
for determining/controling whether subsequent algorithms in the sequence are executed or bypassed

m-fila (approved):

PR #245, opened 2024-10-02
change python formatter to ruff

m-fila:
BEGINRELEASENOTES
- Changed python formatter to ruff

ENDRELEASENOTES

Changing black to ruff and updating the config to use target "py311" as in key4hep/k4geo#398 and key4hep/EDM4hep#370


jmcarcell:
We'll change to ruff once we have a common pre-commit; I'm going to be lazy and close this one since there are conflicts already.

PR #247, opened 2024-10-09
Use bit manipulation with the UniqueIDGenSvc and add a check for repeated numbers

jmcarcell:
BEGINRELEASENOTES
- Overhaul the `UniqueIDGenSvc`:
  - Use bit manipulation with the `UniqueIDGenSvc` which is about 1000 times faster than the string manipulation that was being done
  - Add a check for the case where an ID is repeated - trying to catch the case where the event number and run number is always the same, then the same random number sequence will be generated.

ENDRELEASENOTES

The difference between the bit manipulation versus strings is so big that even when adding the check it will be faster than before. I checked that this PR doesn't change the hash that is returned.


k4FWCore/components/UniqueIDGenSvc.cpp
```diff
@@ -18,27 +18,47 @@
...
+  }
+  for (size_t i = 0; i < bits32; i++) {
+    combined_bits[i + bits32 + bitsSizeT] = event_num_bits[i];
+  }
+  for (size_t i = 0; i < bits32; i++) {
+    combined_bits[i + bitsSizeT] = run_num_bits[i];
+  }
+  for (size_t i = 0; i < bitsSizeT; i++) {
+    combined_bits[i] = name_bits[i];
+  }
```
tmadlener, line 49: Is there a reason to set the bits in this semi reversed order? I think it would be more readable if we reverse the order of loops here, because then all the bits would be set in order. Ideally there would be a way to get rid of the loops entirely, but I don't see anything really obvious apart from bit masking which is not really more readable, IMHO.

k4FWCore/components/UniqueIDGenSvc.cpp
```diff
@@ -18,27 +18,47 @@
...
-  return std::hash<std::bitset<bits64 + bits32 + bits32 + bitsSizeT>>{}(combined_bits);
+  auto                        hash = std::hash<std::bitset<bits64 + bits32 + bits32 + bitsSizeT>>{}(combined_bits);
+  std::lock_guard<std::mutex> lock(m_mutex);
+  if (m_uniqueIDs.contains(hash)) {
+    warning() << "Event number " << evt_num << ", run number " << run_num << " and algorithm name \"" << name
+              << "\" have already been used. Please check the uniqueness of the event number, run number and name."
+              << endmsg;
+  } else {
+    m_uniqueIDs.insert(hash);
+  }
```
tmadlener, line 59:
```suggestion
  bool inserted = false;
  {
    std::lock_guard<std::mutex> lock(m_mutex);
    std::tie(std::ignore, inserted),  = m_uniqueIDs.insert(hash);
  }
  if (!inserted) {
    warning() << "Event number " << evt_num << ", run number " << run_num << " and algorithm name \"" << name
              << "\" have already been used. Please check the uniqueness of the event number, run number and name."
              << endmsg;
  }
```
This way we avoid holding the lock any longer than is necessary and the double lookup inside the set since we`insert` will only insert the value if it is not already present.

k4FWCore/components/UniqueIDGenSvc.cpp
```diff
@@ -49,13 +49,15 @@ size_t UniqueIDGenSvc::getUniqueID(uint32_t evt_num, uint32_t run_num, const std
   }
 
   auto                        hash = std::hash<std::bitset<bits64 + bits32 + bits32 + bitsSizeT>>{}(combined_bits);
-  std::lock_guard<std::mutex> lock(m_mutex);
-  if (m_uniqueIDs.contains(hash)) {
+  bool inserted = false;
+  {
+    std::lock_guard<std::mutex> lock(m_mutex);
+    std::tie(std::ignore, inserted),  = m_uniqueIDs.insert(hash);
```
tmadlener, line 55:
```suggestion
    std::tie(std::ignore, inserted) = m_uniqueIDs.insert(hash);
```
Ah sorry, I messed up the original suggestion. This should work.

k4FWCore/components/UniqueIDGenSvc.cpp
jmcarcell: The order keeps unchanged the current IDs, I reversed it now. For masking they all would need to have the same width so the bitsets would need to be converted to the width of the final one; it is probably very similar in terms of what actually happens internally.

andresailer: Does the reversing the order now mean that the IDs are no longer the same?

tmadlener: IIUC, the order of the bits (i.e. the resulting bitfield) is still the same. My main complaint was about how the bits are set, because the loops started setting bits from `N - 64` to `N`, then did `N - 64 -32` to `N - 64`, etc.

jmcarcell: No, the order is the same, the loops are setting the bits from least significant to most significant, the only thing I did after this suggestion was to reverse the order of the loops in the code.

andresailer: Ok, thanks for clarifying!

k4FWCore/components/UniqueIDGenSvc.cpp
```diff
@@ -18,27 +18,49 @@
...
 
-  return std::hash<std::bitset<bits64 + bits32 + bits32 + bitsSizeT>>{}(combined_bits);
+  auto hash     = std::hash<std::bitset<bits64 + bits32 + bits32 + bitsSizeT>>{}(combined_bits);
+  bool inserted = false;
+  {
+    std::lock_guard<std::mutex> lock(m_mutex);
+    std::tie(std::ignore, inserted) = m_uniqueIDs.insert(hash);
+  }
+  if (!inserted) {
+    warning() << "Event number " << evt_num << ", run number " << run_num << " and algorithm name \"" << name
```
andresailer, line 58: Maybe this should be fatal, with an option to turn it into a warning? Nobody reads warnings.

BrieucF: Is it possible/likely to enter this condition 'by chance' while everything is actually sane? If not then I would be in favor

andresailer: Possible hash collision, but it shouldn't be likely. And if it happens too often, we have a problem we should know about.

jmcarcell: I changed the default to throw whenever the same ID is asked, with an option not to if needed. Indeed if this is buried in thousands of lines of logs it's not very useful. This will throw if someone runs the same algorithm multiple times with the same name; for example, if digitizing multiple collections with the same digitizer and running it multiple times.

PR #248, opened 2024-10-09
Remove redundant inline

jmcarcell:
BEGINRELEASENOTES
- Remove redundant inline, `constexpr` implies `inline`

ENDRELEASENOTES


m-fila:
As far as I understand the `inline` here is not redundant as it's templated, non-static, non-member, constexpr object declaration. And it's in header so it's potentially useful and at least not harmful

jmcarcell:
I agree it doesn't hurt. But I checked and the same code (hashes of .so files match in Release and Debug builds) is produced with and without `inline` so indeed it seems to be redundant. I'd be happy to merge this just to have this closed.

PR #252, opened 2024-10-28
add EventCounter algorithm

### LINKED ISSUE #219: Alternatives for `EventCounter` removed with`GaudiAlg`
A number of repositories uses the `EventCounter` that was part of the deprecated `GaudiAlg`:
- key4hep/k4RecTracker
- HEP-FCC/k4RecCalorimeter
- HEP-FCC/FCCDIRAC 
- HEP-FCC/fcc-tutorials
- HEP-FCC/FCCSW

The [`EventCounter`](https://gitlab.cern.ch/gaudi/Gaudi/-/blob/8dfa672f75a29371e0c02b7820989a0dc1bd689c/GaudiAlg/src/lib/EventCounter.cpp) algorithm itself is simple as it just increments its internal counter on each execution and prints the counter's value on finalize and heartbeat on execution:

- In case the heartbeat printout on execution is desired then a replacement functional algorithm using internal atomic counter could be rather easily written. It should be possible to get the same behavior as in `EventCounter`
- In case the printout only at the finalize is sufficient then there is a newer mechanism in form of `Gaudi::Accumulator::Counter` and `MessageSvcSink` (`MessageSvcSink` prints the values hold in each counter at the finalize). With it, it's possible to:
  - add `Gaudi::Accumulator::Counter` member to some algorithms that should be counted (for instance the input/output algorithms or filters) and remove the `EventCounter` from affected repositories
  - write another `EventCounter` algorithm that does just the counting (can be copied from [`GaudiTestSuite`](https://gitlab.cern.ch/gaudi/Gaudi/-/blob/8dfa672f75a29371e0c02b7820989a0dc1bd689c/GaudiTestSuite/src/FunctionalAlgorithms/EventCounter.cpp))

The solutions involving a separate algorithm doing counting would require to put it in sequential sequencer when used with the `AvalancheSchedulerSvc` (no data dependencies declared so the algorithm could be executed in any order with the respect to the other algorithms)

Alternatively, the information about the event number as seen from a scheduler can be accesses with `EventContext`

m-fila:
BEGINRELEASENOTES
- Added thread-safe`EventCounter` algorithm

ENDRELEASENOTES

Adding thread-safe version of `EventCounter` algorithm. The algorithm will print current event number (as seen by the scheduler) depending on the frequency. Counting the total number of events passed through the algorithm is done with `Gaudi::Accumulators::Counter` - add `MessageSvcSink` in order to get a printout about all the Gaudi counters, for instance:
```txt
26: ApplicationMgr       INFO Application Manager Initialized successfully
26: ApplicationMgr       INFO Application Manager Started successfully
26: EventCounter         INFO Processing event 0
26: EventCounter         INFO Processing event 2
26: EventCounter         INFO Number of counters : 1
26:  |    Counter                                      |     #     |    sum     | mean/eff^* | rms/err^*  |     min     |     max     |
26:  | "count"                                         |         4 |
26: SilentEventCounter   INFO Number of counters : 1
26:  |    Counter                                      |     #     |    sum     | mean/eff^* | rms/err^*  |     min     |     max     |
26:  | "count"                                         |         4 |
26: ApplicationMgr       INFO Application Manager Stopped successfully
26: SilentEventCounter   INFO Processed 4 events
26: EventCounter         INFO Processed 4 events
26: ApplicationMgr       INFO Application Manager Finalized successfully
26: ApplicationMgr       INFO Application Manager Terminated successfully
```

Closes #219

### DIFF
--- a/k4FWCore/components/EventCounter.cpp
+++ b/k4FWCore/components/EventCounter.cpp
@@ -31,6 +31,11 @@ public:
     }
   }
 
+  StatusCode finalize() override {
+    info() << "Processed " << m_count.value() << " events" << endmsg;
+    return Consumer::finalize();
+  }
+
 private:
   mutable Gaudi::Accumulators::Counter<> m_count{this, "count"};
   Gaudi::Property<unsigned int>          m_frequency{this, "Frequency", 1, "How often to print the event number"};
--- a/k4FWCore/components/EventCounter.cpp
+++ b/k4FWCore/components/EventCounter.cpp
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2014-2024 Key4hep-Project.
+ *
+ * This file is part of Key4hep.
+ * See https://key4hep.github.io/key4hep-doc/ for further info.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Gaudi/Accumulators.h"
+#include "Gaudi/Functional/Consumer.h"
+
+class EventCounter final : public Gaudi::Functional::Consumer<void(const EventContext&)> {
+public:
+  using Consumer::Consumer;
+
+  void operator()(const EventContext& ctx) const override {
+    ++m_count;
+    if ((m_frequency > 0) && (ctx.evt() % m_frequency == 0)) {
+      info() << "Processing event " << ctx.evt() << endmsg;
+    }
+  }
+
+private:
+  mutable Gaudi::Accumulators::Counter<> m_count{this, "count"};
+  Gaudi::Property<int>                   m_frequency{this, "Frequency", 1, "How often to print the event number"};
+};
+
+DECLARE_COMPONENT(EventCounter)
--- a/k4FWCore/components/EventCounter.cpp
+++ b/k4FWCore/components/EventCounter.cpp
@@ -26,14 +26,14 @@ public:
 
   void operator()(const EventContext& ctx) const override {
     ++m_count;
-    if ((m_frequency > 0) && (ctx.evt() % m_frequency == 0)) {
+    if (m_frequency && (ctx.evt() % m_frequency == 0)) {
       info() << "Processing event " << ctx.evt() << endmsg;
     }
   }
 
 private:
   mutable Gaudi::Accumulators::Counter<> m_count{this, "count"};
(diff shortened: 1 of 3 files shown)


k4FWCore/components/EventCounter.cpp
```diff
@@ -0,0 +1,39 @@
...
+  void operator()(const EventContext& ctx) const override {
+    ++m_count;
+    if ((m_frequency > 0) && (ctx.evt() % m_frequency == 0)) {
+      info() << "Processing event " << ctx.evt() << endmsg;
+    }
+  }
+
+private:
+  mutable Gaudi::Accumulators::Counter<> m_count{this, "count"};
+  Gaudi::Property<int>                   m_frequency{this, "Frequency", 1, "How often to print the event number"};
```
jmcarcell, line 36:
```suggestion
  Gaudi::Property<unsigned int>                   m_frequency{this, "Frequency", 1, "How often to print the event number"};
```
This could be an `unsigned int` since it's always positive. Then the check above can be slightly shorter with `if (m_frequency && ...`.

m-fila: thanks, fixed

jmcarcell (commented):
I have a few more comments: the original used to have some messages at `initialize()` and `finalize()` and also the format was slightly different using `EventCounter::` in the name. I think this is a matter of opinion. Maybe @BrieucF and @giovannimarchiori can comment if you like how it is right now (note the counter tables are opt-in, which I think some people will find confusing, why is there a sum, rms, etc).

giovannimarchiori:
Thanks a lot for this PR!
Would it be possible to add at finalize a simple printout of the form

EventCounter         INFO Processed XX events

before the (optional) printout of the tables?
How can one configure the EventCounter to avoid showing the tables BTW?

Thanks again!

m-fila:
Thanks for the comments. Added finalize - the message seems to be printed after the tables (updated example in the description)

> How can one configure the EventCounter to avoid showing the tables BTW?

To avoid printing tables for counters don't add `MessageSvcSink` to `ExtSvc` (or change the `MessageSvcSink` property to skip printing tables for all the counters)

giovannimarchiori:
Thanks @m-fila ,
that looks good to me!
   Giovanni

test/k4FWCoreTest/options/TestEventCounter.py
```diff
@@ -0,0 +1,58 @@
...
+
+counter_1 = EventCounter("SilentEventCounter", Frequency=0, OutputLevel=INFO)
+counter_2 = EventCounter("EventCounter", Frequency=2, OutputLevel=INFO)
+
+
+mgr = ApplicationMgr(
+    TopAlg=[counter_1, counter_2],
+    EvtSel="NONE",
+    EvtMax=4,
+    ExtSvc=[whiteboard, "Gaudi::Monitoring::MessageSvcSink"],
```
jmcarcell, line 55:
```suggestion
    # Add "MessageSvcSink" for a table at the end with a statistical summary of the counter data
    # ExtSvc=[whiteboard, "Gaudi::Monitoring::MessageSvcSink"],
    ExtSvc=[whiteboard],
```
Since this is going to be copied and pasted I think by default removing the table is less confusing and `finalize()` already tells how many events were processed.

jmcarcell (approved):

PR #253, opened 2024-10-28
Fix DataHandle::get() for non-collection types and remove dead code

jmcarcell:
BEGINRELEASENOTES
- Fix DataHandle::get() for non-collection types and remove dead code. A `reinterpret_cast` was changed to `static_cast` in https://github.com/key4hep/k4FWCore/pull/250 that made it fail at compile time for non-collection types. Now the `static_cast` is properly wrapped around an `if constexpr` and code around it has been deleted since it looked impossible to trigger.

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/include/k4FWCore/DataHandle.h
+++ b/k4FWCore/include/k4FWCore/DataHandle.h
@@ -22,10 +22,7 @@
 #include "k4FWCore/DataWrapper.h"
 #include "k4FWCore/PodioDataSvc.h"
 
-#include "Gaudi/Algorithm.h"
 #include "GaudiKernel/DataObjectHandle.h"
-
-#include "edm4hep/Constants.h"
 
 #include <GaudiKernel/AnyDataWrapper.h>
 #include <type_traits>
@@ -68,8 +65,6 @@ public:
 
 private:
   ServiceHandle<IDataProviderSvc> m_eds;
-  bool                            m_isGoodType{false};
-  bool                            m_isCollection{false};
   T*                              m_dataPtr;
 };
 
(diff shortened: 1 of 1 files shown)


k4FWCore/include/k4FWCore/DataHandle.h
```diff
@@ -108,40 +103,28 @@ DataHandle<T>::DataHandle(const std::string& descriptor, Gaudi::DataHandle::Mode
...
+  // We wrap it inside constexpr because if the handle has a type that is not a collection
+  // then the static_cast will fail at compile time
+  if constexpr (std::is_base_of_v<podio::CollectionBase, T>) {
+    auto ptr = static_cast<AnyDataWrapper<std::unique_ptr<podio::CollectionBase>>*>(dataObjectp);
+    return static_cast<const T*>(ptr->getData().get());
   }
-  std::string msg("Could not retrieve product " + DataObjectHandle<DataWrapper<T>>::pythonRepr());
-  throw GaudiException(msg, "wrong product name", StatusCode::FAILURE);
+  std::string errorMsg("The type provided for " + DataObjectHandle<DataWrapper<T>>::pythonRepr() +
+                       " is different from the one of the object in the store.");
```
andresailer, line 126: can we print the type we have in the obj store as well?

jmcarcell: I believe this is also impossible to trigger right now but removing it would mean there is possibly a path where the function does not return and that would crash if that ever happens. I added a `typeid` which tells you a mangled type name which is better than nothing.

jmcarcell:
`k4Gen` needs this to compile. Changes needed for `k4RecCalorimeter` after this are here: https://github.com/key4hep/k4RecTracker/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen.

PR #257, opened 2024-11-20
fix: require podio 1.0.1

wdconinc:
BEGINRELEASENOTES
- Require podio 1.0.1 in CMake

ENDRELEASENOTES

Since https://github.com/key4hep/k4FWCore/pull/233, k4FWCore requires podioIO as a target. That's only available as https://github.com/AIDASoft/podio/commit/6eefdc42bcdc26652f20af398429826c0bfff54f, or tag v01-00-01. This change should be propagated to key4hep-spack as well, and k4FWCore should be tested in CI against oldest allowed versions. Tagging @tmadlener @jmcarcell.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,11 +22,7 @@ project(k4FWCore)
 
 find_package(ROOT COMPONENTS RIO Tree REQUIRED)
 find_package(Gaudi REQUIRED)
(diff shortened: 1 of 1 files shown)


tmadlener (approved):
Thanks. Apparently this slipped through until now.

Issue #260, opened 2024-11-29
Bug in saving vectors to configuration_metadata ?


giovannimarchiori:
I am running k4run in the nightly and then inspecting the configuration_metadata of the produced output file. Some of the properties - typically vectors - seem not to be filled.
Example: for this algorithm (https://github.com/HEP-FCC/k4RecCalorimeter/blob/main/RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.h) that I configure with 
```
    augmentClusterAlg = AugmentClustersFCCee("Augment" + outputClusters,
                                                 inClusters=clusterAlg.clusters.Path,
                                                 outClusters="Augmented" + clusterAlg.clusters.Path,
                                                 systemIDs=[4],
                                                 systemNames=["EMB"],
                                                 numLayers=[ecalBarrelLayers],
                                                 readoutNames=[inputReadouts["ecalBarrel"]],
                                                 layerFieldNames=["layer"],
                                                 thetaRecalcWeights=[ecalBarrelThetaWeights],
                                                 # do_photon_shapeVar=runPhotonIDTool,
                                                 do_photon_shapeVar=True,  # we want these variables to train the photon ID BDT
                                                 do_widthTheta_logE_weights=logEWeightInPhotonID,
                                                 OutputLevel=INFO
                                                 )
        TopAlg += [augmentClusterAlg]

```
I see in the configuration_metadata:
```
, AugmentEMBCaloClusters.AuditAlgorithms = "";
, AugmentEMBCaloClusters.AuditExecute = "True";
, AugmentEMBCaloClusters.AuditFinalize = "";
, AugmentEMBCaloClusters.AuditInitialize = "";
, AugmentEMBCaloClusters.AuditReinitialize = "";
, AugmentEMBCaloClusters.AuditRestart = "";
, AugmentEMBCaloClusters.AuditStart = "";
, AugmentEMBCaloClusters.AuditStop = "";
, AugmentEMBCaloClusters.Blocking = "";
, AugmentEMBCaloClusters.Cardinality = "";
, AugmentEMBCaloClusters.Enable = "";
, AugmentEMBCaloClusters.ErrorMax = "";
, AugmentEMBCaloClusters.ExtraInputs = "";
, AugmentEMBCaloClusters.ExtraOutputs = "";
, AugmentEMBCaloClusters.FilterCircularDependencies = "";
, AugmentEMBCaloClusters.MonitorService = "";
[...]

PR #262, opened 2024-12-04
Write the configuration_metadata at initialize() in the Writer

jmcarcell:
to avoid having properties that are deleted because other algorithms are wrapped in a Sequencer and this Sequencer is deleted before `finalize()` is called for Writer. This is what is happening in https://github.com/key4hep/k4FWCore/issues/260 and why some variables are not being saved. The service that is used to get the properties holds references to the properties, so if the properties are deleted, well, apparently they will become empty. This doesn't happen when not using the IOSvc and Writer because the algorithms are not wrapped inside a Sequencer, so the `finalize()` sequence is called for all of them before any of them is deleted.

BEGINRELEASENOTES
- Write the configuration_metadata at initialize() in the Writer, to avoid having properties that are deleted because other algorithms are wrapped in a Sequencer and this Sequencer is deleted before `finalize()` is called for Writer.
- Add tests for different combinations of old/functional algorithms and using PodioOutput or IOSvc and Writer.

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/components/Writer.cpp
+++ b/k4FWCore/components/Writer.cpp
@@ -82,10 +82,6 @@ public:
       return StatusCode::FAILURE;
     }
 
-    return StatusCode::SUCCESS;
-  }
-
-  StatusCode finalize() override {
     podio::Frame config_metadata_frame;
 
     //// prepare job options metadata ///////////////////////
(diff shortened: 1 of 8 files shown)


jmcarcell:
Merging this today if there aren't any comments

andresailer (approved):
Thanks!

m-fila (approved):
lgtm

giovannimarchiori:
Thanks @jmcarcell !

PR #265, opened 2024-12-09
Remove the check for TrackerHit3D from edm4hep

jmcarcell:
BEGINRELEASENOTES
- Remove the check for TrackerHit3D from edm4hep

ENDRELEASENOTES


tmadlener (approved):

jmcarcell:
I updated the required version of EDM4hep to 0.99 since `k4FWCore` doesn't build with previous versions (with or without this PR).

PR #269, opened 2024-12-18
Improve the documentation about extending the parser for k4run

jmcarcell:
BEGINRELEASENOTES
- Improve the documentation about extending the parser for k4run

ENDRELEASENOTES

Some minor syntax changes and a comment about the `[0]` that I typically forget when writing from memory.

### DIFF
--- a/doc/k4run-args.md
+++ b/doc/k4run-args.md
@@ -18,7 +18,8 @@ limitations under the License.
 -->
 # Adding custom arguments to `k4run`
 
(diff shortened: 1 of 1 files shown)


m-fila (approved):
Could you also add usage example from cli? For instance something like

> The example steering file can then be executed as follows:
```
k4run example_steering_file.py --trackingOnly --some-other-args 
```

jmcarcell:
Done

PR #270, opened 2024-12-18
Use CMake fixtures for tests that require other tests

jmcarcell:
BEGINRELEASENOTES
- Use CMake fixtures for tests that require other tests
  - Tests will now run only if all the setup tests pass, otherwise they will be skipped

ENDRELEASENOTES

There is no downside to using fixtures except needing to add names for the setup - required groups. Also `ctest -R <name>` will run also the tests that `<name>` depends on, but this can be overwritten with `-FA ".*"`. The advantage is failing faster. `DEPENDS` only signals ordering.

```
 4/65 Test #30: FunctionalProducer ................................***Failed    2.63 sec
Failed test dependencies: FunctionalProducer
 5/65 Test #36: FunctionalMTFile ..................................***Not Run   0.00 sec
Failed test dependencies: FunctionalProducer
 6/65 Test #34: FunctionalFile ....................................***Not Run   0.00 sec
Failed test dependencies: FunctionalProducer
 7/65 Test #59: FunctionalFileCLI .................................***Not Run   0.00 sec
Failed test dependencies: FunctionalProducer
 8/65 Test #58: FunctionalFileTooShort ............................***Not Run   0.00 sec
Failed test dependencies: FunctionalProducer
 9/65 Test #65: GaudiFunctional ...................................***Not Run   0.00 sec
Failed test dependencies: FunctionalProducer
```


jmcarcell:
If there aren't any comments I'll merge today

PR #271, opened 2024-12-19
Remove the unnecessary r-value reference

tmadlener:
The frame is move only in any case

BEGINRELEASENOTES
- Remove an unnecessary r-value reference qualifier for sinking the metadata Frame into the `MetadataSvc`.

ENDRELEASENOTES

Mainly because I saw [this lightning talk](https://www.youtube.com/watch?v=iH86aQrzIKU) yesterday and this seems to be a case where it's unnecessary.

(We have this in quite a few places, where the *move-only* property of our types already ensures that a `std::move` on the calling site is necessary if not using a temporary)

### DIFF
--- a/k4FWCore/components/MetadataSvc.cpp
+++ b/k4FWCore/components/MetadataSvc.cpp
@@ -45,6 +45,6 @@ StatusCode MetadataSvc::finalize() { return Service::finalize(); }
 
 const podio::Frame* MetadataSvc::getFrame() const { return m_frame.get(); }
 podio::Frame*       MetadataSvc::getFrame() { return m_frame.get(); }
-void MetadataSvc::setFrame(podio::Frame&& frame) { m_frame = std::make_unique<podio::Frame>(std::move(frame)); }
+void MetadataSvc::setFrame(podio::Frame frame) { m_frame = std::make_unique<podio::Frame>(std::move(frame)); }
 
 DECLARE_COMPONENT(MetadataSvc)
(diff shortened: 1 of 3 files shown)


tmadlener:
Yes, indeed. Title should be r-value not universal reference

m-fila (approved):
Sounds very reasonable, points taken :smile: 
I couldn't find a clang-tidy check to flag such pattern

Core guidelines rules [F.16](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const) and [F.18](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f18-for-will-move-from-parameters-pass-by-x-and-stdmove-the-parameter)

jmcarcell:
> Sounds very reasonable, points taken 😄 I couldn't find a clang-tidy check to flag such pattern
> 
> Core guidelines rules [F.16](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const) and [F.18](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f18-for-will-move-from-parameters-pass-by-x-and-stdmove-the-parameter)

I think it's only F.18 that applies here, and the key part is for why making this change is just simplicity:
> Passing by value does generate one extra (cheap) move operation, but prefer simplicity and clarity first.

Issue #272, opened 2024-12-19
Discussion about ParticleID metadata


tmadlener:
This came up in a discussion with @BrieucF and @saracreates. Before diving too far into an implementation, I would like to see what people prefer from an interface / usage point of view. 

TL;DR: We need an easy way to access ParticleID related metadata that makes using the `PIDHandler` in algorithms easily possible. Main question: Do we want another service specifically for this or can we adapt the metadata svc?

------
EDM4hep has some utilities to conveniently write particle id related metadata, e.g. algorithm and parameter names associated with a given `ParticleIDCollection`. These utilites are defined in [`edm4hep/utils/ParticleIDUtils.h`](https://github.com/key4hep/EDM4hep/blob/main/utils/include/edm4hep/utils/ParticleIDUtils.h). On a very basic level they are just some fancy wrappers around Fame parameters that use a naming convention to store the collection level metadata. If the metadata follow these conventions it is easy to construct a `edm4hep::utils::PIDHandler` which can then be used to more easily retrieve the metadata, but also to directly access the correct algorithm and / or specific parameters for an algorithm. The utilities all effectively assume that they have access to a Frame, but the API can also be used with the `edm4hep::utils::ParticleIDMeta` to pass the necessary metadata in.

It is currently already possible to mimic what the utilities are doing by just carefully putting the necessary metadata into the `MetadataSvc`. However, this is rather error prone, and I would ideally like to hide this from the users.

There are several possible approaches
- We could change the `IMetadataSvc` interface to also have a `put` overload that takes an `edm4hep::utils::ParticleIDMeta` and also add either a template specialization to `get` or add a `getAlgoInfo` (or similar) that returns a `std::optional<edm4hep::utils::ParticleIDMeta>`.
- We could introduce an entirely new service, e.g. `ParticleIDMetadataSvc` that offers only this specific functionality.
- We open the `IMetadataSvc` to give access to it's internal Frame and people could use the EDM4hep utilities directly.

For me the third option is not really on the table, since we explicitly wanted to avoid the direct access to the Frame.

[...]

PR #273, opened 2024-12-20
Add the possibility to set `edm4hep::utils::ParticleIDMeta` via the `IMetadataSvc`

tmadlener:
BEGINRELEASENOTES
- Make it possible to use `edm4hep::utils::ParticleIDMeta` with the `MetadataSvc`
  - Add template specializations for `get` and `put` that defer to the corresponding utility calls in EDM4hep
  - Add tests to ensure that metadata is indeed usable this way with the utilities in EDM4hep  

ENDRELEASENOTES

This would be option 1 sketched out in #272 

- [x] Small test that checks that metadata is also usable outside of framework
- [x] Needs https://github.com/key4hep/EDM4hep/pull/395

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ find_package(ROOT COMPONENTS RIO Tree REQUIRED)
 find_package(Gaudi REQUIRED)
 find_package(podio 1.0.1 REQUIRED)
 find_package(EDM4HEP 0.99 REQUIRED)
+find_package(fmt REQUIRED)
 
 include(cmake/Key4hepConfig.cmake)
 
@@ -45,7 +46,6 @@ add_subdirectory(k4FWCore)
 add_subdirectory(k4Interface)
 add_subdirectory(python)
 if(BUILD_TESTING)
-  find_package(fmt REQUIRED)
   add_subdirectory(test/k4FWCoreTest)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,7 @@ add_subdirectory(k4FWCore)
 add_subdirectory(k4Interface)
 add_subdirectory(python)
 if(BUILD_TESTING)
+  find_package(fmt REQUIRED)
   add_subdirectory(test/k4FWCoreTest)
 endif()
 
--- a/test/k4FWCoreTest/CMakeLists.txt
+++ b/test/k4FWCoreTest/CMakeLists.txt
@@ -206,3 +206,10 @@ add_test_with_env(ParticleIDMetadataFramework options/ExampleParticleIDMetadata.
 add_custom_command(TARGET k4FWCoreTestPlugins POST_BUILD
                    COMMAND ${CMAKE_COMMAND} -E copy_directory
                    ${PROJECT_SOURCE_DIR}/python/k4FWCore/ ${PROJECT_BINARY_DIR}/k4FWCore/genConfDir/k4FWCore)
+
+
+add_executable(check_ParticleIDOutputs src/check_ParticleIDOutputs.cpp)
+target_link_libraries(check_ParticleIDOutputs PRIVATE podio::podioIO EDM4HEP::edm4hep EDM4HEP::utils fmt::fmt)
+add_test(NAME check_ParticleIDOutputs COMMAND check_ParticleIDOutputs example_with_particleids.root)
+set_test_env(check_ParticleIDOutputs)
+set_tests_properties(check_ParticleIDOutputs PROPERTIES DEPENDS ParticleIDMetadataFramework)
--- a/test/k4FWCoreTest/CMakeLists.txt
+++ b/test/k4FWCoreTest/CMakeLists.txt
@@ -195,6 +195,8 @@ add_test_with_env(FunctionalProducerRNTuple options/ExampleFunctionalProducerRNT
 add_test_with_env(FunctionalTTreeToRNTuple options/ExampleFunctionalTTreeToRNTuple.py PROPERTIES FIXTURES_REQUIRED ProducerFile ADD_TO_CHECK_FILES)
 add_test_with_env(GaudiFunctional options/ExampleGaudiFunctional.py PROPERTIES FIXTURES_REQUIRED ProducerFile ADD_TO_CHECK_FILES)
 
+add_test_with_env(ParticleIDMetadataFramework options/ExampleParticleIDMetadata.py)
+
 
 # The following is done to make the tests work without installing the files in
 # the installation directory. The k4FWCore in the build directory is populated by
--- a/test/k4FWCoreTest/src/check_ParticleIDOutputs.cpp
+++ b/test/k4FWCoreTest/src/check_ParticleIDOutputs.cpp
@@ -1,0 +1,85 @@
+#include <edm4hep/ReconstructedParticleCollection.h>
+#include <edm4hep/utils/ParticleIDUtils.h>
+
+#include <podio/Reader.h>
+
+#include <fmt/format.h>
+#include <fmt/ranges.h>
+
+#include <algorithm>
+
+void checkPIDForAlgo(const edm4hep::utils::PIDHandler& pidHandler, const edm4hep::ReconstructedParticle& reco,
+                     const edm4hep::utils::ParticleIDMeta& pidMeta, const int paramIndex) {
+  auto maybePID = pidHandler.getPID(reco, pidMeta.algoType());
+  if (!maybePID) {
+    throw std::runtime_error(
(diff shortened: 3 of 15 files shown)


tmadlener:
I have added a "small" test case that runs similar checks as in the functional algorithm as a standalone application (effectively hard-coding the expected parameter and algorithm names). I have done it in c++ because the PIDHandler is not yet really usable in python (see https://github.com/key4hep/EDM4hep/issues/396)

test/k4FWCoreTest/CMakeLists.txt
```diff
@@ -204,3 +206,10 @@ add_test_with_env(GaudiFunctional options/ExampleGaudiFunctional.py PROPERTIES F
 add_custom_command(TARGET k4FWCoreTestPlugins POST_BUILD
                    COMMAND ${CMAKE_COMMAND} -E copy_directory
                    ${PROJECT_SOURCE_DIR}/python/k4FWCore/ ${PROJECT_BINARY_DIR}/k4FWCore/genConfDir/k4FWCore)
+
+
+add_executable(check_ParticleIDOutputs src/check_ParticleIDOutputs.cpp)
+target_link_libraries(check_ParticleIDOutputs PRIVATE podio::podioIO EDM4HEP::edm4hep EDM4HEP::utils fmt::fmt)
```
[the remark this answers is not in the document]
tmadlener: Yes, good idea. I was relying on `find_package(Gaudi)` bringing it in. Should I put it into the top level `CMakeLists.txt` or in the one in `tests`? Currently only the tests seem to be using fmtlib, but that might change(?)

test/k4FWCoreTest/src/check_ParticleIDOutputs.cpp
```diff
@@ -0,0 +1,95 @@
...
+    throw std::runtime_error(
+        fmt::format("Could not retrieve the correct parameter value for param {} (expected {}, actual {})",
+                    pidMeta.paramNames[paramIndex], paramIndex * 0.5f, paramVal));
+  }
+}
+
+bool checkAlgoMetadata(const edm4hep::utils::ParticleIDMeta& pidMeta, const std::string& algoName,
+                       const std::vector<std::string>& paramNames) {
+  if (pidMeta.algoName != algoName) {
+    fmt::print(
```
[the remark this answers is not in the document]
tmadlener: Ah yes, good catch. This was a mess. Fixed to no longer use exceptions.

tmadlener:
Since I touched the `ExampleFunctionalProducerMultiple` to also create `ReconstructedParticle`s, I had to fix a few other tests to also make them pass. The alternative would be to introduce a dedicated producer for the ParticleID metadata tests. I am not sure if it's worth it, but it can obviously be done.

test/k4FWCoreTest/src/check_ParticleIDOutputs.cpp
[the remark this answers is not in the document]
tmadlener: Added it to the top level one, but behind the `BUILD_TESTING` check.

test/k4FWCoreTest/CMakeLists.txt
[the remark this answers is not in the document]
jmcarcell: Well with https://github.com/key4hep/k4FWCore/pull/275 it will change, so it may be at the top level anyway.

test/k4FWCoreTest/src/components/ExampleParticleIDConsumer.cpp
```diff
@@ -0,0 +1,154 @@
...
+              << endmsg;
+    }
+
+    return StatusCode::SUCCESS;
+  }
+
+  void operator()(const edm4hep::ParticleIDCollection& pidColl1, const edm4hep::ParticleIDCollection& pidColl2,
+                  const edm4hep::ReconstructedParticleCollection& recos) const {
+    auto pidHandler = edm4hep::utils::PIDHandler::from(pidColl1, pidColl2);
+    pidHandler.addMetaInfos(m_pidMeta1, m_pidMeta2);
```
tmadlener, line 111: This requires a new version of EDM4hep. Should I go for a version that works with 0.99? Otherwise we need to bump the EDM4hep version, I think. Even if this is only a test.

jmcarcell: Up to you. But actually 0.99.1 in EDM4hep has the problem that https://github.com/key4hep/EDM4hep/pull/379 is not merged in and this causes issues when trying to build on top of EDM4hep 0.99.1 since it won't find `nlohmann_json`.

jmcarcell:
I'll give this another look and update https://github.com/key4hep/k4MarlinWrapper/pull/216 to use this to see if anything is missing between today and tomorrow.

test/k4FWCoreTest/src/components/ExampleParticleIDProducer.cpp
```diff
@@ -0,0 +1,71 @@
...
+  StatusCode initialize() final {
+    m_pidMeta            = {m_pidAlgoName, m_pidParamNames};
+    std::string collname = outputLocations("ParticleIDCollection")[0];
+    k4FWCore::putParameter(collname, m_pidMeta);
+    return StatusCode::SUCCESS;
+  }
+
+  edm4hep::ParticleIDCollection operator()(const edm4hep::ReconstructedParticleCollection& recos) const {
+    auto pidColl = edm4hep::ParticleIDCollection{};
+    for (const auto r : recos) {
```
jmcarcell, line 46:
```suggestion
    for (const auto& r : recos) {
```

test/k4FWCoreTest/src/components/ExampleParticleIDConsumer.cpp
```diff
@@ -0,0 +1,154 @@
...
+      error() << fmt::format("Could not get a parameter index for {} (got {}) or {} (got {})",
+                             m_paramOfInterest1.value(), m_paramIndex1, m_paramOfInterest2.value(), m_paramIndex2)
+              << endmsg;
+    }
+
+    return StatusCode::SUCCESS;
+  }
+
+  void operator()(const edm4hep::ParticleIDCollection& pidColl1, const edm4hep::ParticleIDCollection& pidColl2,
+                  const edm4hep::ReconstructedParticleCollection& recos) const {
```
jmcarcell, line 109:
```suggestion
                  const edm4hep::ReconstructedParticleCollection& recos) const override {
```

test/k4FWCoreTest/src/components/ExampleParticleIDConsumer.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include "k4FWCore/Consumer.h"
+#include "k4FWCore/MetadataUtils.h"
+
+#include "edm4hep/ParticleIDCollection.h"
+#include "edm4hep/utils/ParticleIDUtils.h"
+
+#include <Gaudi/Property.h>
+
+#include "fmt/format.h"
+#include "fmt/ranges.h"
```
jmcarcell, line 30:
```suggestion
```
Only `fmt::format` is being used

tmadlener: No, that's a false positive (negative?) from llvm. `pidMeta.paramNames` is a `std::vector` which requires `fmt/ranges.h`.

CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ find_package(ROOT COMPONENTS RIO Tree REQUIRED)
 find_package(Gaudi REQUIRED)
 find_package(podio 1.0.1 REQUIRED)
 find_package(EDM4HEP 0.99 REQUIRED)
+find_package(fmt REQUIRED)
```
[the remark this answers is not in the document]
tmadlener: Indeed it should.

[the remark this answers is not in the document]
tmadlener: And it now is.

test/k4FWCoreTest/src/components/ExampleParticleIDConsumer.cpp
tmadlener: I have changed it such that CI still passes on the `release`s.

jmcarcell:
Merge?

As a note once we stop using Ubuntu 22.04 (because there are already builds for Ubuntu 24.04, https://github.com/key4hep/key4hep-spack/pull/700) we can use `<format>`; anyway Gaudi doesn't support GCC 11 anymore...

m-fila (approved):

tmadlener:
> Merge?

Yes, this is done from my point of view, unless someone still has comments.

PR #276, opened 2025-01-31
Clean up the function that reads inputs in functional algorithms

jmcarcell:
Changes extracted from #275

BEGINRELEASENOTES
- Clean up the function that reads inputs in functional algorithms by defining an alias for the EDM4hep type and using that when possible. Remove `const` from this type and add it where needed.
- Change `std::remove_cvref_t` to `std::remove_cv_t` since there the former is not needed.

ENDRELEASENOTES


k4FWCore/include/k4FWCore/FunctionalUtils.h
```diff
@@ -127,24 +127,22 @@ namespace k4FWCore {
...
+          auto inputMap     = std::vector<const EDM4hepType*>();
           for (auto& handle : std::get<Index>(handles)) {
-            if constexpr (std::is_same_v<EDM4hepType, const podio::CollectionBase*>) {
-              inputMap.push_back(&get(handle, thisClass, Gaudi::Hive::currentContext()));
-            } else {
-              podio::CollectionBase* in = handle.get()->get();
-              inputMap.push_back(static_cast<EDM4hepType*>(in));
-            }
+            podio::CollectionBase* in = handle.get()->get();
+            inputMap.push_back(static_cast<const EDM4hepType*>(in));
```
jmcarcell, line 137: I think this answers the question about the CollectionMerger @tmadlener, it seems that both branches are equivalent. I'm not sure how, I guess when I added it it wasn't compiling for the Merger but it's fine to be removed.

tmadlener (commented):
I think removing the `if constexpr` might break the `EDM4hepType::typeName` trick for printing the name in #275, but I am not entirely sure (plus we can try to fix it there if necessary).

You said you need changes also for something else, but this seems to be a pure refactoring at the moment. Will those come in another PR, or this one?

jmcarcell:
It will be another PR, unless I find something else right now this one is ready.

tmadlener (approved):

PR #279, opened 2025-02-03
Fix algorithm properties related to `eventNumber` and `runNumber` to deduce their type

m-fila:
BEGINRELEASENOTES
- Fix algorithm properties related to `eventNumber` and `runNumber` to deduce their type

ENDRELEASENOTES

This is follow-up to #278, that changes types of remaining properties that are related to `eventNumer` and `runNumber`.
This is mostly to be consistent and avoid implicit conversions or a potential footgun with adding negative event-offset to unsigned event number

### DIFF
--- a/test/k4FWCoreTest/src/components/ExampleEventHeaderConsumer.cpp
+++ b/test/k4FWCoreTest/src/components/ExampleEventHeaderConsumer.cpp
@@ -55,11 +55,12 @@ struct ExampleEventHeaderConsumer final : k4FWCore::Consumer<void(const edm4hep:
     }
   }
 
-  Gaudi::Property<int> m_runNumber{this, "runNumber", 0, "The expected run number"};
-  Gaudi::Property<int> m_eventNumberOffset{this, "eventNumberOffset", 0,
-                                           "The event number offset where events will start counting from"};
-  // Since the type of the runNumber changed recently in EDM4hep, use its type
+  // Since the type of the runNumber and eventNumber changed recently in EDM4hep, deduce their type
   // instead of hardcoding it to avoid warnings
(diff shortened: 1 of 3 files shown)


test/k4FWCoreTest/src/components/ExampleEventHeaderConsumer.cpp
```diff
@@ -55,11 +55,12 @@ struct ExampleEventHeaderConsumer final : k4FWCore::Consumer<void(const edm4hep:
     }
   }
 
-  Gaudi::Property<int> m_runNumber{this, "runNumber", 0, "The expected run number"};
-  Gaudi::Property<int> m_eventNumberOffset{this, "eventNumberOffset", 0,
-                                           "The event number offset where events will start counting from"};
-  // Since the type of the runNumber changed recently in EDM4hep, use its type
+  // Since the type of the runNumber and eventNumber changed recently in EDM4hep, deduce their type
   // instead of hardcoding it to avoid warnings
+  Gaudi::Property<decltype(std::declval<edm4hep::EventHeader>().getEventNumber())> m_runNumber{
```
jmcarcell, line 60: I don't understand how there is a comparison above without a warning, when it is comparing `int` and `uint32_t`. Maybe because it goes through the `Property::operator!=` but the types are preserved there: https://gitlab.cern.ch/gaudi/Gaudi/-/blob/master/install/include/Gaudi/Property.h#L206

m-fila: Yeah, I too think there should be a warning. Still didn't figure out why there is none. Going through `.value()` gives a warning

jmcarcell (approved):

PR #280, opened 2025-02-03
Add support for reading collections stored as DataWrapper<podio::CollectionBase> in functional algorithms

jmcarcell:
This is needed for https://github.com/key4hep/k4MarlinWrapper/pull/216, because the collections that the Marlin wrapper produces are stored as `DataWrapper<podio::CollectionBase>` (see https://github.com/key4hep/k4MarlinWrapper/blob/main/k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp#L83), which we don't use anywhere else.

BEGINRELEASENOTES
- Make functional algorithms read `DataWrapper<podio::CollectionBase>` 
- Add a test producing a collection in a `DataWrapper<podio::CollectionBase>` and reading it in a functional algorithm

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/include/k4FWCore/FunctionalUtils.h
+++ b/k4FWCore/include/k4FWCore/FunctionalUtils.h
@@ -175,10 +175,9 @@ namespace k4FWCore {
               DataObject*       p;
               IDataProviderSvc* svc = thisClass->evtSvc();
               svc->retrieveObject("/Event/" + std::get<Index>(handles)[0].objKey(), p).ignore();
-              // This is how Gaudi::Algorithms will save collections through the DataHandle
+              // This is how Gaudi::Algorithms saves collections through the DataHandle
               const auto* wrp = dynamic_cast<const DataWrapper<EDM4hepType>*>(p);
-              // This is how the Marlin wrapper will save collections when converting
-              // from LCIO to EDM4hep
+              // This is how the Marlin wrapper saves collections when converting from LCIO to EDM4hep
               const auto* marlinWrp = dynamic_cast<const DataWrapper<podio::CollectionBase>*>(p);
               if (!wrp && !marlinWrp) {
                 throw GaudiException(thisClass->name(),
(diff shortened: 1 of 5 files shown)


tmadlener:
I suppose it's not easily possible to make the MarlinWrapper to store them differently?

jmcarcell:
I don't think so. The "normal" way would be `DataWrapper<edm4hep::MCParticle>` for example, but then one needs to know the type and then it's not possible to go over all the collections as it is done now. There would have to be a templated function with all the types and an `if - else` construct to call the correct template depending on the type, more or less like it is done in k4EDM4hep2LcioConv. And the change here is quite small.

tmadlener (approved):

k4FWCore/include/k4FWCore/FunctionalUtils.h
```diff
@@ -175,14 +175,23 @@ namespace k4FWCore {
               DataObject*       p;
               IDataProviderSvc* svc = thisClass->evtSvc();
               svc->retrieveObject("/Event/" + std::get<Index>(handles)[0].objKey(), p).ignore();
+              // This is how Gaudi::Algorithms will save collections through the DataHandle
```
jmcarcell, line 178:
```suggestion
              // This is how Gaudi::Algorithms saves collections through the DataHandle
```

k4FWCore/include/k4FWCore/FunctionalUtils.h
```diff
@@ -175,14 +175,23 @@ namespace k4FWCore {
...
               const auto* wrp = dynamic_cast<const DataWrapper<EDM4hepType>*>(p);
-              if (!wrp) {
+              // This is how the Marlin wrapper will save collections when converting
+              // from LCIO to EDM4hep
```
jmcarcell, line 181:
```suggestion
              // This is how the Marlin wrapper saves collections when converting from LCIO to EDM4hep
```

PR #282, opened 2025-02-04
Use typeid instead of EDM4hep::typeName

jmcarcell:
Introduced in https://github.com/key4hep/k4FWCore/pull/275, same as https://github.com/key4hep/k4FWCore/pull/277, see https://github.com/AIDASoft/podio/issues/731.

BEGINRELEASENOTES
- Use typeid instead of `EDM4hep::typeName` since not every EDM4hep type has `::typeName`.

ENDRELEASENOTES

I forgot to change it in the PR, I even thought about it :(
I'm in the process of adding some tests with Links, it seems we have none and we are blind to this from k4FWCore.


jmcarcell:
Merging this to build the nightlies for tomorrow, I'll add the tests I mention later.

PR #283, opened 2025-02-04
Add a Link collection to the tests

jmcarcell:
we didn't have any, and things like https://github.com/key4hep/k4FWCore/pull/282 wouldn't happen because that would have not compiled.

BEGINRELEASENOTES
- Add a Link collection to the tests

ENDRELEASENOTES

### DIFF
--- a/test/k4FWCoreTest/src/components/ExampleFunctionalConsumerMultiple.cpp
+++ b/test/k4FWCoreTest/src/components/ExampleFunctionalConsumerMultiple.cpp
@@ -118,8 +118,8 @@ struct ExampleFunctionalConsumerMultiple final
     for (size_t j = 0; j < links.size(); j++) {
       if (links[j].getFrom() != recos[j] || links[j].getTo() != particles[j]) {
         std::stringstream error;
-        error << "Wrong data in links collection, link" << j << " expected " << recos[0] << ", " << particles[j]
-              << " got " << links[j].getFrom() << ", " << links[j].getTo();
+        error << "Wrong data in links collection, link" << j << " expected " << recos[j].id() << ", " << particles[j].id()
+              << " got " << links[j].getFrom().id() << ", " << links[j].getTo().id();
         throw std::runtime_error(error.str());
       }
     }
--- a/test/k4FWCoreTest/src/components/ExampleFunctionalConsumerMultiple.cpp
+++ b/test/k4FWCoreTest/src/components/ExampleFunctionalConsumerMultiple.cpp
@@ -118,8 +118,8 @@ struct ExampleFunctionalConsumerMultiple final
     for (size_t j = 0; j < links.size(); j++) {
       if (links[j].getFrom() != recos[j] || links[j].getTo() != particles[j]) {
         std::stringstream error;
-        error << "Wrong data in links collection, link" << j << " expected " << recos[j].id() << ", " << particles[j].id()
-              << " got " << links[j].getFrom().id() << ", " << links[j].getTo().id();
+        error << "Wrong data in links collection, link" << j << " expected " << recos[j].id() << ", "
+              << particles[j].id() << " got " << links[j].getFrom().id() << ", " << links[j].getTo().id();
         throw std::runtime_error(error.str());
       }
     }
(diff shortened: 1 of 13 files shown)


test/k4FWCoreTest/src/components/ExampleFunctionalConsumerMultiple.cpp
```diff
@@ -107,6 +114,15 @@ struct ExampleFunctionalConsumerMultiple final
             << tracks[0].getChi2() << ", " << tracks[0].getNdf();
       throw std::runtime_error(error.str());
     }
+
+    for (size_t j = 0; j < links.size(); j++) {
+      if (links[j].getFrom() != recos[j] || links[j].getTo() != particles[j]) {
+        std::stringstream error;
+        error << "Wrong data in links collection, link" << j << " expected " << recos[0] << ", " << particles[j]
+              << " got " << links[j].getFrom() << ", " << links[j].getTo();
```
tmadlener, line 122: Doesn't need to happen in this PR (or maybe not at all?), but we are starting to get a wild mix of `fmt::format` and `std::stringstream` for formatting error messages. I think we should settle for one for Key4hep and then try to use that consistently.

test/k4FWCoreTest/src/components/ExampleFunctionalTransformerRuntimeCollectionsMultiple.cpp
```diff
@@ -180,8 +192,50 @@ struct ExampleFunctionalTransformerRuntimeCollectionsMultiple final
...
+                               "");
+    }
+    for (auto& recoColl : recoVec) {
+      auto coll = edm4hep::ReconstructedParticleCollection();
+      // if (recoColl->size() != 5 || recoColl->at(1).getPDG() != 1) {
+      //   std::stringstream error;
+      //   // error << "Wrong data in reco collection, expected 0, 1, 2, 3, 4 got " << coll[0].getPDG() << ", "
+      //   //       << coll[1].getPDG() << ", " << coll[2].getPDG() << ", " << coll[3].getPDG() << ", " << coll[4].getPDG();
+      //   throw std::runtime_error(error.str());
+      // }
```
tmadlener, line 206: Is this a left over from developing? I think if this passes for all the tests we could just uncomment it and leave it in. Similar below.

test/k4FWCoreTest/src/components/ExampleFunctionalConsumerMultiple.cpp
jmcarcell: Yes, let's go for `std::format` once that Ubuntu 22 is dropped which will be soon since Gaudi hasn't been supporting GCC 11 for a while and it always has to be patched.

tmadlener: I think we can go for `fmt` for now in any case. Gaudi uses that internally quite heavily, in that case we wouldn't need to rely on compiler support as much.

test/k4FWCoreTest/src/components/ExampleFunctionalTransformerRuntimeCollectionsMultiple.cpp
jmcarcell: They are uncommented now

test/k4FWCoreTest/src/components/ExampleFunctionalConsumerMultiple.cpp
jmcarcell: But then we change it once to change it later to `std::format`, and if we were to follow the systems that are supported by Gaudi we would already have compiler support for it (they only build for GCC >= 13 and Clang >= 16, same as the LCG stacks) so I would perefer to keep the `std::stringstream`s and do the changes at once in the not so distant future.

test/k4FWCoreTest/src/components/ExampleFunctionalTransformerRuntimeCollectionsMultiple.cpp
```diff
@@ -180,8 +192,50 @@ struct ExampleFunctionalTransformerRuntimeCollectionsMultiple final
...
+      }
+      for (const auto& reco : *recoColl) {
+        coll.push_back(reco.clone());
+      }
+      recoVecOut.emplace_back(std::move(coll));
+    }
+
+    if (linkVec.size() != 3) {
+      throw std::runtime_error("Wrong size of the link vector, expected 3, got " + std::to_string(linkVecOut.size()) +
+                               "");
```
tmadlener, line 218:
```suggestion
      throw std::runtime_error("Wrong size of the link vector, expected 3, got " + std::to_string(linkVecOut.size()));
```

tmadlener (approved):
Minor formatting thing, otherwise looks good to me.

PR #259, opened 2024-11-26
CI: move lcg test to alma9

andresailer:
BEGINRELEASENOTES
- CI: move lcg test to alma9, fixes #258

ENDRELEASENOTES

PR #286, opened 2025-02-12
Add a dev3 workflow using the key4hep-build action

jmcarcell:
in place of https://github.com/key4hep/k4FWCore/pull/259

BEGINRELEASENOTES
- Add a dev3 workflow using the key4hep-build action

ENDRELEASENOTES

Other tests with LCG stacks will come in the "main" build workflow, but since there isn't any for dev3 planned and k4FWCore is a good place to test dev3 for ROOT issues I added the workflow.


jmcarcell:
Pre-commit should be fixed...

PR #287, opened 2025-02-13
Use the same integer widths in `UniqueIDGenSvc` as in `EventHeader`

m-fila:
BEGINRELEASENOTES
- Use the same event and run type in `UniqueIDGenSvc` as in `edm4hep::EventHeader`. Add generating id directly from `edm4hep::EventHeader`. **This changes the ID values !**

ENDRELEASENOTES

Calling `getUniqueID` with arguments taken from `edm4hep::EventHeader` is the most common usage of that function. Until https://github.com/key4hep/EDM4hep/pull/398 the function arguments and `EventHeader` had different signedness.
After that PR they have the same signedness but different widths, with the function arguments being shorter - which means there is a huge range of valid `EventHeaders` that will get the same ID due to an implicit cast

I propose to deduct the correct types from `EventHeader` as in #278 and #279. I also propose to add an overload taking `EventHeader` directly since this is the most common usage and avoids a possible issue with manually assigning wrong type.

**This changes the ID values !**

### DIFF
--- a/k4FWCore/components/UniqueIDGenSvc.cpp
+++ b/k4FWCore/components/UniqueIDGenSvc.cpp
@@ -18,38 +18,39 @@
  */
 #include "UniqueIDGenSvc.h"
 
-#include <cstdint>
+#include <cstddef>
 #include <stdexcept>
 #include <string>
 
+constexpr size_t event_num_digits = std::numeric_limits<UniqueIDGenSvc::event_num_t>::digits;
+constexpr size_t run_num_digits   = std::numeric_limits<UniqueIDGenSvc::run_num_t>::digits;
+constexpr size_t seed_digits      = std::numeric_limits<UniqueIDGenSvc::seed_t>::digits;
+constexpr size_t name_digits      = std::numeric_limits<size_t>::digits;
+
 UniqueIDGenSvc::UniqueIDGenSvc(const std::string& name, ISvcLocator* svcLoc) : base_class(name, svcLoc) {}
 
-constexpr size_t bits32    = std::numeric_limits<uint32_t>::digits;
-constexpr size_t bits64    = std::numeric_limits<uint64_t>::digits;
-constexpr size_t bitsSizeT = std::numeric_limits<size_t>::digits;
+size_t UniqueIDGenSvc::getUniqueID(event_num_t evt_num, run_num_t run_num, const std::string& name) const {
+  std::bitset<seed_digits>      seed_bits      = m_seed.value();
+  std::bitset<event_num_digits> event_num_bits = evt_num;
+  std::bitset<run_num_digits>   run_num_bits   = run_num;
+  std::bitset<name_digits>      name_bits      = std::hash<std::string>{}(name);
 
-size_t UniqueIDGenSvc::getUniqueID(uint32_t evt_num, uint32_t run_num, const std::string& name) const {
-  std::bitset<bits64>    seed_bits      = this->m_seed.value();
-  std::bitset<bits32>    event_num_bits = evt_num, run_num_bits = run_num;
-  size_t                 str_hash  = std::hash<std::string>{}(name);
-  std::bitset<bitsSizeT> name_bits = str_hash;
+  std::bitset<seed_digits + event_num_digits + run_num_digits + name_digits> combined_bits;
 
-  std::bitset<bits64 + bits32 + bits32 + bitsSizeT> combined_bits;
-
-  for (size_t i = 0; i < bitsSizeT; i++) {
+  for (size_t i = 0; i < name_digits; i++) {
     combined_bits[i] = name_bits[i];
   }
-  for (size_t i = 0; i < bits32; i++) {
-    combined_bits[i + bitsSizeT] = run_num_bits[i];
+  for (size_t i = 0; i < run_num_digits; i++) {
+    combined_bits[i + name_digits] = run_num_bits[i];
   }
-  for (size_t i = 0; i < bits32; i++) {
-    combined_bits[i + bits32 + bitsSizeT] = event_num_bits[i];
+  for (size_t i = 0; i < event_num_digits; i++) {
+    combined_bits[i + run_num_digits + name_digits] = event_num_bits[i];
   }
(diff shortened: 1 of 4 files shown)


k4FWCore/components/UniqueIDGenSvc.cpp
```diff
@@ -18,38 +18,39 @@
...
-#include <cstdint>
+#include <cstddef>
 #include <stdexcept>
 #include <string>
 
-UniqueIDGenSvc::UniqueIDGenSvc(const std::string& name, ISvcLocator* svcLoc) : base_class(name, svcLoc) {}
+constexpr size_t event_num_digits = std::numeric_limits<UniqueIDGenSvc::event_num_t>::digits;
+constexpr size_t run_num_digits   = std::numeric_limits<UniqueIDGenSvc::run_num_t>::digits;
+constexpr size_t seed_digits      = std::numeric_limits<UniqueIDGenSvc::seed_t>::digits;
+constexpr size_t name_digits      = std::numeric_limits<size_t>::digits;
```
tmadlener, line 28: Should these be in a namespace rather than the global one?

k4Interface/include/k4Interface/IUniqueIDGenSvc.h
```diff
@@ -32,8 +34,13 @@
...
 
 struct GAUDI_API IUniqueIDGenSvc : extend_interfaces<IInterface> {
+  using event_num_t = decltype(std::declval<edm4hep::EventHeader>().getEventNumber());
+  using run_num_t   = decltype(std::declval<edm4hep::EventHeader>().getRunNumber());
   DeclareInterfaceID(IUniqueIDGenSvc, 2, 0);
-  virtual size_t getUniqueID(uint32_t evt_num, uint32_t run_num, const std::string& name) const = 0;
+  virtual size_t getUniqueID(event_num_t evt_num, run_num_t run_num, const std::string& name) const = 0;
+  size_t         getUniqueID(const edm4hep::EventHeader& evt_header, const std::string& name) const {
+    return getUniqueID(evt_header.getEventNumber(), evt_header.getRunNumber(), name);
+  }
```
tmadlener, line 43: Maybe also an overload that takes the `edm4hep::EventHeaderCollection` and simply uses the first element for calling this?

k4FWCore/components/UniqueIDGenSvc.h
```diff
@@ -35,11 +35,13 @@
  */
 class UniqueIDGenSvc : public extends<Service, IUniqueIDGenSvc> {
 public:
+  using seed_t = uint64_t;
+
```
tmadlener, line 39: Should this potentially also live in the `IUniqueIDGenSvc` like the `event_num_t` and `run_num_t`?

k4Interface/include/k4Interface/IUniqueIDGenSvc.h
m-fila: added

k4FWCore/components/UniqueIDGenSvc.cpp
m-fila: They're in cpp file so they shouldn't leak. I don't know if there is any benefit in making them members, otherwise what namespace should these go into?

tmadlener: Ah, yes. I missed the fact that they are in the cpp file and not the header. All good then. sorry for the noise.

k4FWCore/components/UniqueIDGenSvc.h
tmadlener: That makes sense.

k4FWCore/components/UniqueIDGenSvc.cpp
m-fila: No problem, thanks for the review :slightly_smiling_face:

k4Interface/include/k4Interface/IUniqueIDGenSvc.h
```diff
@@ -32,8 +34,17 @@
  */
 
 struct GAUDI_API IUniqueIDGenSvc : extend_interfaces<IInterface> {
+  using event_num_t = decltype(std::declval<edm4hep::EventHeader>().getEventNumber());
+  using run_num_t   = decltype(std::declval<edm4hep::EventHeader>().getRunNumber());
   DeclareInterfaceID(IUniqueIDGenSvc, 2, 0);
-  virtual size_t getUniqueID(uint32_t evt_num, uint32_t run_num, const std::string& name) const = 0;
+  virtual size_t getUniqueID(event_num_t evt_num, run_num_t run_num, const std::string& name) const = 0;
```
jmcarcell, line 40:
```suggestion
  virtual size_t getUniqueID(const event_num_t evt_num, const run_num_t run_num, const std::string& name) const = 0;
```
Also for the derived class

k4FWCore/components/UniqueIDGenSvc.h
```diff
@@ -35,11 +35,13 @@
...
   UniqueIDGenSvc(const std::string& name, ISvcLocator* svcLoc);
-  size_t getUniqueID(uint32_t evt_num, uint32_t run_num, const std::string& name) const override;
+  size_t getUniqueID(event_num_t evt_num, run_num_t run_num, const std::string& name) const override;
```
jmcarcell, line 41:
```suggestion
  size_t getUniqueID(conts event_num_t evt_num, const run_num_t run_num, const std::string& name) const override;
```

jmcarcell (approved):
Note that this now adds the possibility of changing the IDs by changing fields in EDM4hep, which wasn't happening before. Maybe this is never important, as someone trying to reproduce some runs should use the same versions of EDM4hep and k4FWCore.

jmcarcell:
I'll defer merging to @tmadlener since he never approved it :smile:

tmadlener (approved):
(Assuming that #294 will be addressed in other PRs).

m-fila:
> (Assuming that #294 will be addressed in other PRs).

Certainly. Thanks!

PR #288, opened 2025-02-14
update UniqueIDGenSvc docs and add example usage in functional algorithm

### LINKED ISSUE #285: Document how to use UniqueGenIDSVc in functional world

m-fila:
BEGINRELEASENOTES
- Update`UniqueIDGenSvc`documentation and add example usage with a functional algorithm

ENDRELEASENOTES

Adding a test with example functional algorithm using `UniqueIDSvc`. Selected parts of the algorithm are then copied to the documentation.

After key4hep/key4hep-doc#91 the page will be also included in the main documentation

Closes #285

### DIFF
--- a/doc/uniqueIDGen.md
+++ b/doc/uniqueIDGen.md
@@ -68,6 +68,9 @@ Then, use the service during execution (`execute` member function for the algori
 
 ```cpp
 StatusCode ExampleAlgorithm::execute(const EventContext&) const {
+  // Instead of passing event number and run number separately,
+  // an overload accepting EventHeader or EventHeaderCollection can be also used.
+  StatusCode sc = m_service->getUniqueID(1, 2, name());
   m_service->getUniqueID(1, 2, name());
 }
 ```
@@ -112,10 +115,10 @@ During algorithm execution the `EventHeaderCollection` can be used to obtain a u
 ```cpp
 public:
   podio::UserDataCollection<double> operator()(const edm4hep::EventHeaderCollection& evtHeader) const final {
-    const auto evt = evtHeader[0];
-
-    // obtain unique value
-    auto uid = m_uniqueIDSvc->getUniqueID(evt.getEventNumber(), evt.getRunNumber(), name());
+    // obtain unique value from the first header in a collection
+    auto uid = m_uniqueIDSvc->getUniqueID(evtHeader, name());
+    // or from a given EventHeader object
+    // auto uid = m_uniqueIDSvc->getUniqueID(evtHeader[0], name());
 
     // seed TRandom3 or some other PRNG of your choice
     auto prng = TRandom3(uid);
--- a/doc/uniqueIDGen.md
+++ b/doc/uniqueIDGen.md
@@ -16,20 +16,40 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
-# How to use UniqueIDGenSvc
 
-A Service to generate unique identifiers can be used with the following inputs:
-- Event number, Run number and Algorithm name
-- Seed number: set in the options python file
+# Seeding algorithm PRNG
 
-To set a seed for the UniqueIDGenSvc, add the following to an options file:
+Correctly seeding pseudo-random number generators (PRNG) is important to ensuring reproducible and good-quality randomness. A common pattern is to reinitialize each algorithm’s PRNG with a unique seed for each processed event.
+
+## UniqueIDGenSvc service
+
+The `UniqueIDGenSvc` Gaudi service can generate unique `size_t` values that can be used as PRNG seeds, based on an event number, run number, and algorithm.
+
+The service can be included in a steering file by:
 
 ```python
 from Configurables import UniqueIDGenSvc
-UniqueIDGenSvc().Seed = 987654321
+uidgen_svc = UniqueIDGenSvc()
 ```
 
-Declare the service in the header file:
+The service can be configured with a central seed from which the unique values will be derived. This seed can be configured with the `Seed` property:
+
+::::{tab-set}
+:::{tab-item} Python
+```python
+uidgen_svc.Seed = 987654321
+```
+:::
+:::{tab-item} CLI
+```sh
+k4run <steering-file> --UniqueIDGenSvc.Seed 987654321
+```
+:::
+::::
+
+## Using UniqueIDGenSvc in algorithms
+
+To use the `UniqueIDGenSvc` in a Gaudi algorithm first include the header file and declare a member variable:
 
 ```cpp
 #include <k4FWCore/IUniqueIDGenSvc.h>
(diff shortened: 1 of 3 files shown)


doc/uniqueIDGen.md
```diff
@@ -16,38 +16,114 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+The service can be configured with a central seed from which the unique values will be derived. This seed can be configured with the `Seed` property:
+
+::::{tab-set}
+:::{tab-item} Python
+```python
+uidgen_svc.Seed = 987654321
+```
+:::
+:::{tab-item} CLI
```
andresailer, line 43: Is it possible to directly link to the CLI part of this tabbing?

doc/uniqueIDGen.md
```diff
@@ -16,38 +16,114 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-- Event number, Run number and Algorithm name
-- Seed number: set in the options python file
+# Seeding algorithm PRNG
 
-To set a seed for the UniqueIDGenSvc, add the following to an options file:
+Correctly seeding pseudo-random number generators (PRNG) is important to ensuring reproducible and good-quality randomness. A common pattern is to reinitialize each algorithm’s PRNG with a unique seed for each processed event.
+
+## UniqueIDGenSvc service
+
+The `UniqueIDGenSvc` Gaudi service can generate unique `size_t` values that can be used as PRNG seeds, based on an event number, run number, and algorithm.
```
tmadlener, line 26:
```suggestion
The `UniqueIDGenSvc` Gaudi service can generate unique `size_t` values that can be used as PRNG seeds, based on an event number, run number, and algorithm name.
```
Just to be precise here.

doc/uniqueIDGen.md
```diff
@@ -16,38 +16,114 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+::::{tab-set}
+:::{tab-item} Python
+```python
+uidgen_svc.Seed = 987654321
+```
+:::
+:::{tab-item} CLI
+```sh
+k4run <steering-file> --UniqueIDGenSvc.Seed 987654321
```
tmadlener, line 45: This assumes that the user doesn't give it another name. It will work with the code presented in here, but I am never sure whether we should highlight this nuance.

doc/uniqueIDGen.md
```diff
@@ -16,38 +16,114 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+During algorithm execution the `EventHeaderCollection` can be used to obtain a unique value from the `UniqueIDSvc`, then the value can be used as a seed for PRNG. In this example the PRNG is then used to generate a random value and push it to an output collection.
+
+```cpp
+public:
+  podio::UserDataCollection<double> operator()(const edm4hep::EventHeaderCollection& evtHeader) const final {
+    const auto evt = evtHeader[0];
+
+    // obtain  unique value
+    auto uid = m_uniqueIDSvc->getUniqueID(evt.getEventNumber(), evt.getRunNumber(), name());
```
tmadlener, line 118: With the changes in #287 this could be simplified a bit more? Or do we want to show this explicitly and simply mention the convenience overload in the text? Simply for the examples' sake I would have a preference for the former, but i am also fine with an explicit approach here.

doc/uniqueIDGen.md
m-fila: I tried to do it in the same style as with IOSvc docs in which we refer just to `--IOSvc.whatever`

I think the ability to name the service differently can be quite problematic for users, because the algorithms will still look for a service by a name we have hardcoded (and possibly initialize it if not found and that's enabled)

doc/uniqueIDGen.md
m-fila: Yes, I'd like to promote using `EventHeader` or `EventHeaderCollection` overload as they are also less error prone to implicit narrowing casts

tmadlener: so probably better to not mention the option here at all.

doc/uniqueIDGen.md
tmadlener: So probably not mention the possibility here at all, to not cause confusion.

doc/uniqueIDGen.md
andresailer: Thanks for investigating! I think I would propose to have this linkable somehow?

tmadlener:
pre-commit fails due to https://github.com/cvmfs-contrib/github-action-cvmfs/issues/33


However, I have no idea where it pulls the cache from, as I just removed all the caches in k4FWCore, but it still gets them from somewhere:

https://github.com/key4hep/k4FWCore/actions/runs/13332703762/job/37326017166#step:3:23

andresailer:
> pre-commit fails due to [cvmfs-contrib/github-action-cvmfs#33](https://github.com/cvmfs-contrib/github-action-cvmfs/issues/33)
> 
> However, I have no idea where it pulls the cache from, as I just removed all the caches in k4FWCore, but it still gets them from somewhere:
> 
> https://github.com/key4hep/k4FWCore/actions/runs/13332703762/job/37326017166#step:3:23

Update cvmfs action to v4?

tmadlener:
> Update cvmfs action to v4 v5?

#289 Let's see if that does the trick. Thanks for the hint there, didn't even realize there were newer versions.

doc/uniqueIDGen.md
```diff
@@ -16,38 +16,114 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 
 ```cpp
-StatusCode SomeGaudiAlgorithm::initialize() {
-  m_service = serviceLocator()->service("UniqueIDGenSvc");
+StatusCode ExampleAlgorithm::initialize() {
+  m_service = service("UniqueIDGenSvc");
 ```
 
-Then use the service during execution:
+Then, use the service during execution:
```
jmcarcell, line 67:
```suggestion
Then, use the service during execution (for algorithms using `Gaudi::Algorithm`):
```
Something like this to point out that this is not always the case?

doc/uniqueIDGen.md
```diff
@@ -16,38 +16,114 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+```
+
+During algorithm execution the `EventHeaderCollection` can be used to obtain a unique value from the `UniqueIDSvc`, then the value can be used as a seed for PRNG. In this example the PRNG is then used to generate a random value and push it to an output collection.
+
+```cpp
+public:
+  podio::UserDataCollection<double> operator()(const edm4hep::EventHeaderCollection& evtHeader) const final {
+    const auto evt = evtHeader[0];
+
+    // obtain  unique value
```
jmcarcell, line 117:
```suggestion
    // obtain unique value
```

doc/uniqueIDGen.md
```diff
@@ -16,38 +16,114 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+    auto uid = m_uniqueIDSvc->getUniqueID(evt.getEventNumber(), evt.getRunNumber(), name());
+
+    // seed TRandom3 or some other PRNG of your choice
+    auto prng = TRandom3(uid);
+
+    auto coll = podio::UserDataCollection<double>();
+    coll.push_back(a);
```
jmcarcell, line 124:
```suggestion
    coll.push_back(prng.Rndm());
```

doc/uniqueIDGen.md
m-fila: Using overload with `EventHeaderCollection` now

doc/uniqueIDGen.md
```diff
@@ -16,38 +16,117 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+```cpp
+public:
+  podio::UserDataCollection<double> operator()(const edm4hep::EventHeaderCollection& evtHeader) const final {
+    // obtain unique value from the first header in a collection
+    auto uid = m_uniqueIDSvc->getUniqueID(evtHeader, name());
+    // or from a given EventHeader object
+    // auto uid = m_uniqueIDSvc->getUniqueID(evtHeader[0], name());
+
+    // seed TRandom3 or some other PRNG of your choice
```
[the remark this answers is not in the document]
andresailer: It probably can't be a class member because the functions called on it are not going to be `const`?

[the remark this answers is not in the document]
m-fila: `thread_local` members must be also `static`. But they can be used in `const` member functions since they don't belong to an instance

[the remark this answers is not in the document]
andresailer: Thanks for the further clarification @m-fila !

I still don't think we should document the RNG as member variable, because people will just find away around having to initialise this RNG for every event...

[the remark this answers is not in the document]
m-fila: I agree, it's probably too easy to abuse

PR #290, opened 2025-02-19
Properly limit the collections that are read

### LINKED ISSUE #150: PodioInput does not limit the collections that are read
The `collections` options of the `PodioInput` does not work as intended, I think. It reads the complete Frame, i.e. all collections, but then effectively just gives access to the ones that are specified (even others would be available). More importantly, in combination with `PodioOutput` all collections that are available on file will be written to the outputfile as well, which is quite unexpected IMHO.

tmadlener:
BEGINRELEASENOTES
- Make the `PodioInput.collections`, `IOSvc.CollectionNames` and `Reader.InputCollections` properties actually work like expected. They now properly limit the collections that are read and no other collections will be available. **This requires building against podio > v1.2`**, otherwiset the current behavior will be used, where collections that are not requested will still be available.

ENDRELEASENOTES

Fixes #150 

- [x] ~documentation for `IOSvc`(?)~ already present
- [x] Needs https://github.com/AIDASoft/podio/pull/739 and one pass of the nightlies to be sure the tests for the `k4DataSvc` version are actually checking that this works (otherwise they are just skipped).

Especially for the `IOSvc` + `Reader` side of things there are quite a few handles to change which collections should be read and they have to be kept somewhat consistent. I think we should try to find a way to harmonize that a bit, but I am not entirely sure which way is the one that should be kept. @jmcarcell any preference there, resp. do you see more ways this could be made more robust?


k4FWCore/components/IIOSvc.h
```diff
@@ -39,8 +39,9 @@ class IIOSvc : virtual public IInterface {
    * @brief Read the next event from the input file
    * @return A tuple containing the collections read, the collection names and the frame that owns the collections
    */
-  virtual std::tuple<std::vector<podio::CollectionBase*>, std::vector<std::string>, podio::Frame> next() = 0;
-  virtual std::shared_ptr<std::vector<std::string>> getCollectionNames() const                           = 0;
+  virtual std::tuple<std::vector<podio::CollectionBase*>, std::vector<std::string>, podio::Frame> next()       = 0;
+  virtual std::shared_ptr<std::vector<std::string>> getCollectionNames() const                                 = 0;
+  virtual void                                      setReadingCollectionNames(const std::vector<std::string>&) = 0;
```
jmcarcell, line 44: This is not needed because the IOSvc has the collections that it is going to read in `m_collectionNames`.

k4FWCore/components/Reader.cpp
```diff
@@ -106,22 +108,24 @@ class Reader final : public CollectionPusher {
...
+  ServiceHandle<IIOSvc> m_iosvc{this, "IOSvc", "IOSvc"};
 
   StatusCode initialize() override {
-    if (!iosvc.isValid()) {
+    if (!m_iosvc.isValid()) {
       error() << "Unable to locate IIOSvc interface" << endmsg;
       return StatusCode::FAILURE;
     }
+    debug() << "Initializing IOSvc to read collections: " << m_inputCollections << endmsg;
+    m_iosvc->setReadingCollectionNames(m_inputCollections);
```
jmcarcell, line 121: Not needed, see above

test/k4FWCoreTest/CMakeLists.txt
```diff
@@ -194,6 +195,7 @@ add_test_with_env(FunctionalReadNthEvent options/ExampleFunctionalReadNthEvent.p
 add_test_with_env(FunctionalProducerRNTuple options/ExampleFunctionalProducerRNTuple.py ADD_TO_CHECK_FILES)
 add_test_with_env(FunctionalTTreeToRNTuple options/ExampleFunctionalTTreeToRNTuple.py PROPERTIES FIXTURES_REQUIRED ProducerFile ADD_TO_CHECK_FILES)
 add_test_with_env(GaudiFunctional options/ExampleGaudiFunctional.py PROPERTIES FIXTURES_REQUIRED ProducerFile ADD_TO_CHECK_FILES)
+add_test_with_env(ReadLimitedInputsIOSvc options/readLimitedSetOfCollections.py PROPERTIES FIXTURES_REQUIRED ExampleEventDataFile ADD_TO_CHECK_FILES)
```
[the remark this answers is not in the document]
tmadlener: Done. Now it's `ExampleIOSvcLimitInputCollections.py`

tmadlener:
I have refactored the `fix_properties` slightly. Mainly I have lifted the setup concerning the reader into a `_setup_reader` method that effectively does the same thing as before. However, now there are a few more checks whether `import podio` is really necessary. That will only happen now if `EvtMax == -1` or `CollectionNames` is empty.

The on thing that can become inconsistent now is when a user instantiates their own `Reader` in python and sets a different `InputCollections` on that than they set on `CollectionNames` for the `IOSvc`. I am not sure how likely that scenario is though (we don't even have an example with it AFAICT).

tmadlener:
The problems in the writer should also be fixed now. They have become necessary because the changes to the our python wrapper around the `ApplicationMgr` have started to fill the `InputCollections` of the `Reader` and that in turn means that not all of the collections from the `Frame` get pushed to the TES. However, the `Writer` tried to remove all of the Frame collections unconditionally. I have made the early returns simple `continue`s in the loop and demoted the `error` to `warning` which makes everything work again. I have also added a bit more debug output to the writer since I needed it in any case to figure out what was going on. I think it could be useful for future debugging via logs, but I can also remove it again.

python/k4FWCore/ApplicationMgr.py
```diff
@@ -101,7 +104,16 @@ def fix_properties(self):
                 except IndexError:
                     print("Warning, the events category wasn't found in the input file")
                     raise
-                collections = list(frame.getAvailableCollections())
+                # We have to carry over the desired input collections from the IOSvc
```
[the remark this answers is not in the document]
jmcarcell: This is now done

python/k4FWCore/ApplicationMgr.py
```diff
@@ -37,6 +40,64 @@ class ApplicationMgr:
...
+            return
+
+        collections = iosvc_props["CollectionNames"][0] or None
+        n_events = self._mgr.EvtMax
+
+        if collections and n_events != -1:
+            # We know everything we need
+            logger.info(f"Initializing reader to read {n_events} events and read: {collections}")
+            reader.InputCollections = collections
+            return
```
jmcarcell, line 81:
```suggestion
        if collections:
            reader.InputCollections = collections
            if n_events != -1:
                # We know everything we need
                logger.info(f"Initializing reader to read {n_events} events and read: {collections}")
                return
```
If any collections are passed then set the collections unconditionally, this is how it was being done before. I'll make a comment below.

python/k4FWCore/ApplicationMgr.py
```diff
@@ -75,34 +136,12 @@ def fix_properties(self):
...
-                podio_reader = PodioReader(props[inp][0])
-                if self._mgr.EvtMax == -1:
-                    self._mgr.EvtMax = podio_reader._reader.getEntries("events")
-                try:
-                    frame = podio_reader.get("events")[0]
-                except IndexError:
-                    print("Warning, the events category wasn't found in the input file")
-                    raise
-                collections = list(frame.getAvailableCollections())
-                reader.InputCollections = collections
```
jmcarcell, line 105: This was always running when there was input, otherwise the Reader doesn't have information about which collections is going to read which is needed as a property because of all the stuff (creating the handles that will be used to push data) that happens before `initialize()`.

python/k4FWCore/ApplicationMgr.py
tmadlener: Oh yeah, good catch. I completely missed that. I have added a test that covers this scenario as well now. The failure I get without this fix is also what made me put in the "fixes" for the Writer that are not strictly necessary, because otherwise the writer would try to get non-existent collections.

k4FWCore/components/Writer.cpp
```diff
@@ -220,15 +225,16 @@ class Writer final : public Gaudi::Functional::Consumer<void(const EventContext&
     // deleted by the store (and later deleted by the Frame, triggering a double
     // delete)
     for (const auto& coll : frameCollections) {
+      debug() << "Taking ownership of collection " << coll << " from the IOSvc as it belongs to a Frame" << endmsg;
       DataObject* storeCollection;
       if (m_dataSvc->retrieveObject("/Event/" + coll, storeCollection).isFailure()) {
-        error() << "Failed to retrieve collection " << coll << endmsg;
-        return;
+        warning() << "Failed to retrieve collection " << coll << endmsg;
+        continue;
```
jmcarcell, line 232:
```suggestion
        error() << "Failed to retrieve collection " << coll << endmsg;
        return;
```
Keep it as it is since it doesn't need to be changed and then maybe change in another PR?

k4FWCore/components/Writer.cpp
```diff
@@ -220,15 +225,16 @@ class Writer final : public Gaudi::Functional::Consumer<void(const EventContext&
...
       }
       // We take ownership back from the store
       if (m_dataSvc->unregisterObject(storeCollection).isFailure()) {
-        error() << "Failed to unregister collection " << coll << endmsg;
-        return;
+        warning() << "Failed to unregister collection " << coll << endmsg;
+        continue;
```
jmcarcell, line 237:
```suggestion
        error() << "Failed to unregister collection " << coll << endmsg;
        return;
```

tmadlener: I messed up reverting these somehow. Now they should be back to the previous version.

python/k4FWCore/ApplicationMgr.py
```diff
@@ -37,6 +40,67 @@ class ApplicationMgr:
...
+        podio_reader = PodioReader(iosvc_props[inp][0])
+        if n_events == -1:
+            self._mgr.EvtMax = len(podio_reader.get("events"))
+        if not collections:
+            try:
+                frame = podio_reader.get("events")[0]
+                logger.debug("Using the first frame to determine collections to read")
+                collections = list(frame.getAvailableCollections())
+            except IndexError:
+                print("Warning, the events category wasn't found in the input file")
```
tmadlener, line 99:
```suggestion
                logger.warning("Warning, the events category wasn't found in the input file")
```
Since we use the logging module in other places.

k4FWCore/components/Writer.cpp
```diff
@@ -240,15 +246,16 @@ class Writer final : public Gaudi::Functional::Consumer<void(const EventContext&
 
     std::vector<std::string_view> collectionsToRemove;
     for (const auto& coll : m_collectionsToAdd) {
+      debug() << "Adding collection " << coll << " to the IOSvc Frame" << endmsg;
       DataObject* storeCollection;
       if (m_dataSvc->retrieveObject("/Event/" + coll, storeCollection).isFailure()) {
-        error() << "Failed to retrieve collection " << coll << endmsg;
-        return;
+        warning() << "Failed to retrieve collection " << coll << endmsg;
+        continue;
```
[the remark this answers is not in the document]
tmadlener: OK. Now I think I *really* remember why I did this in the first place. If podio doesn't yet support filling a frame with only a subset of all collections, the Writer crashes (see [this CI run](https://github.com/key4hep/k4FWCore/actions/runs/13583944364/job/37974821491?pr=290)). Changing the early return to a `continue` fixes that issue and we would not need to bump the minimal podio version necessary for k4FWCore.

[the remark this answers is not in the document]
jmcarcell: Ok now I see that is necessary. I don't think it is related to writing only a subset of collections, that has been there for a while and was working before this PR. I think the explanation is related to a mismatch in the collections that are returned with `getAvailableCollections` and the ones that are actually in the store, which is something that couldn't happen before because all the collections were being read. I think the messages should not be warnings then, because imagine for a frame with many collections of which only a single one is read, then there will be a warning for every collection that is not read for every event. Alternatively, if there is a way from obtaining only the collections that are being read from podio (taking into account this doesn't change for different events), then that could solve the issue.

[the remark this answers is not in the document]
tmadlener: TL;DR: Should the message level be `DEBUG`? The only downside I see is that it's easy to miss for users why expected collections are not in the output.

> I think the explanation is related to a mismatch in the collections that are returned with `getAvailableCollections` and the ones that are actually in the store

Yes, exactly. In old (<= 1.2) versions of podio, the Frame will always claim to have all collections from the file available, because there is no way to limit it until https://github.com/AIDASoft/podio/pull/504. So in that case the collection will be missing from the store, and we have to skip that collection. From podio v1.3 it is possible to only get a subset of all collections, so `getAvailableCollections` and the collections in the store will be the same again.

There are two things that would need to happen to make this work (as far as I can see atm):
- We need to get the input collections from `PodioInput`
- We need to additionally also keep track of all collections that are added to the store during processing.

The second one seems pretty intrusive to do here, compared to simply checking whether the collection is there and skipping it if it's not. I think we can make the message level `DEBUG`, at the cost of it not being very obvious to users why collections they expect are not there, unless they turn it on.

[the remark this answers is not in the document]
jmcarcell: I would say a DEBUG message and skipping is fine - this is the only situation where those messages have appeared so I don't think it would be a problem anywhere else. It can to in this PR since it's needed for limiting the collections and not that big of a change.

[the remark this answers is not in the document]
tmadlener: Done.

k4FWCore/components/Writer.cpp
```diff
@@ -220,15 +225,16 @@ class Writer final : public Gaudi::Functional::Consumer<void(const EventContext&
...
-        return;
+        debug() << "Failed to retrieve collection " << coll << " from data service" << endmsg;
+        continue;
       }
       // We take ownership back from the store
       if (m_dataSvc->unregisterObject(storeCollection).isFailure()) {
-        error() << "Failed to unregister collection " << coll << endmsg;
-        return;
+        debug() << "Failed to unregister collection " << coll << " from data service" << endmsg;
+        continue;
```
jmcarcell, line 237:
```suggestion
        error() << "Failed to unregister collection " << coll << " from data service" << endmsg;
        return;
```

This should be independent of the previous one (it's impossible to be here if the collection is not found) so I think this should remain an error, something is probably very wrong if unregistering something that is in the store fails.

Issue #294, opened 2025-02-26
UniqueIDGenSvc uniqueness and reproducibility considerations


m-fila:
Disclaimer: These are considerations in light of what the C++ standard guarantees; in practice, things might just work™

The `UniqueIDGenSvc` is described (and possibly used) as the means to get reproducible pRNG seeds. My concern lies in the reproducibility aspect due to the use of`std::hash` which per the C++ standard guarantees reproducibility only within a single execution of a program - in other words `std::hash` is free to return different values in any scenarios like:
- two different executions of the same program,
- recompilation,
- change of compiler
- different platform.

As a related issue, the`UniqueIDGenSvc` results depend also on the length of `size_t` for which only the minimal length is guaranteed by the standard, which in principle could change the values obtained on different platforms.

The result itself being `size_t` might be also problematic when `UniqueIDGenSvc` result is used as a seed for a generator with fixed width integer seed - a seed might be unique as `size_t` but non-unique after implicit casting to seed type. In practice this should be fine since the collisions shouldn't be often, it's just the users might think they have unique seed but they don't.

The last issue is caching the produced IDs in a set but the set is never cleaned which might exhaust the memory in longer running jobs. Perhaps a property could be used to disable using this cache?

tmadlener:
> The last issue is caching the produced ids in a map but the map is never cleaned which might exhaust the memory in longer running jobs. Perhaps a property could control to disable using this cache

Related to this, I think this is in principle already almost there

We have the `ThrowIfDuplicate` property:
https://github.com/key4hep/k4FWCore/blob/5416eebe468ce5e5092fb4f28bf3323089b1a1a2/k4FWCore/components/UniqueIDGenSvc.h#L45

and that effectively controls whether we only get an error message or a hard failure:
https://github.com/key4hep/k4FWCore/blob/5416eebe468ce5e5092fb4f28bf3323089b1a1a2/k4FWCore/components/UniqueIDGenSvc.cpp#L55-L64

So maybe simply giving that a better name (`CheckDuplicates`?) and only populating the map in case that is enabled could be a solution?

jmcarcell:
Maybe we could have different defaults for debug versus release builds?

tmadlener:
Yes, I think that's a good idea (however, the solution might look in the end).

tmadlener:
For a reproducible hash we are using *MurmurHash3* in a couple of places, e.g.
- in podio for the collection IDs (32 bit hash of the collection name)
[...]

PR #295, opened 2025-02-26
Use a property to control whether duplicate IDs are checked

m-fila:
BEGINRELEASENOTES
- Replace `UniqueIDGenSvc` property `ThrowIfDuplicate` with `CheckDuplicates`. The service will cache ids and check duplicates only if `CheckDuplicates` is set to `true`. It's enabled by default for the Debug builds.

ENDRELEASENOTES

The property will control whether to cache the ids so the long running jobs can disable it to avoid possible memory exhaustion.

As suggested in https://github.com/key4hep/k4FWCore/issues/294#issuecomment-2684538702 the option to control whether in case of duplicate to throw an exception or print error is removed. Now it'll throw an exception. I think it's ok for debugging.

The default value of `CheckDuplicates` depends on a build type and is taken from `NDEBUG`

Part of #294

### DIFF
--- a/k4FWCore/components/UniqueIDGenSvc.h
+++ b/k4FWCore/components/UniqueIDGenSvc.h
@@ -49,11 +49,10 @@ private:
       this, "CheckDuplicates",
   // Default value for release and debug builds
 #ifdef NDEBUG
-      false
+      false,
 #else
-      true
+      true,
 #endif
-      ,
       "Caches obtained ID and throws an exception if a duplicate would be returned"};
 };
 


k4FWCore/components/UniqueIDGenSvc.cpp
```diff
@@ -49,18 +51,16 @@ size_t UniqueIDGenSvc::getUniqueID(uint32_t evt_num, uint32_t run_num, const std
...
+  auto hash = std::hash<std::bitset<bits64 + bits32 + bits32 + bitsSizeT>>{}(combined_bits);
+
+  if (m_checkDuplicates) {
+    bool inserted = [this, hash]() {
+      std::lock_guard<std::mutex> lock(m_mutex);
+      return m_uniqueIDs.insert(hash).second;
+    }();
+    if (!inserted) {
+      throw std::runtime_error(
+          fmt::format("Duplicate event number, run number and algorithm name: {}, {}, {}", evt_num, run_num, name));
```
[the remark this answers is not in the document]
m-fila: How about calling it ID instead of random seed?

tmadlener (approved):

k4FWCore/components/UniqueIDGenSvc.h
```diff
@@ -41,10 +42,17 @@ class UniqueIDGenSvc : public extends<Service, IUniqueIDGenSvc> {
...
+  mutable std::unordered_map<size_t, std::tuple<event_num_t, run_num_t, std::string>, std::identity> m_uniqueIDs;
+  mutable std::mutex                                                                                 m_mutex;
+  Gaudi::Property<bool>                                                                              m_checkDuplicates{
+      this, "CheckDuplicates", m_isDebugBuild,
+      "Caches obtained ID and throws an exception if a duplicate would be returned"};
+#ifdef NDEBUG
+  constexpr static bool m_isDebugBuild = false;
+#else
+  constexpr static bool m_isDebugBuild = true;
+#endif
```
jmcarcell, line 55:
```suggestion
  Gaudi::Property<bool>                                                                              m_checkDuplicates{
      this,
      "CheckDuplicates",
      // Default value for release and debug builds
#ifdef NDEBUG
      false;
#else
      true;
#endif
      "Caches obtained ID and throws an exception if a duplicate would be returned"};
```
Isn't this simpler without the extra `m_isDebugBuild`?

jmcarcell: I don't think there is that much of a difference but there is a bit less of indirection with this approach. Can you give it a try and we see it on the diff?

m-fila: Here it is

k4FWCore/components/UniqueIDGenSvc.h
```diff
@@ -41,10 +42,19 @@ class UniqueIDGenSvc : public extends<Service, IUniqueIDGenSvc> {
...
+  Gaudi::Property<seed_t> m_seed{this, "Seed", {123456789}};
+  Gaudi::Property<bool>   m_checkDuplicates{
+      this, "CheckDuplicates",
+  // Default value for release and debug builds
+#ifdef NDEBUG
+      false
+#else
+      true
+#endif
+      ,
```
jmcarcell, line 56:
```suggestion
#ifdef NDEBUG
      false,
#else
      true,
#endif
```

Issue #240, opened 2024-09-26
WARNING: duplicated property name


giovannimarchiori:
When recompiling a nightly with k4FWCore locally downloaded, I noticed this

[ 90%] Linking CXX shared module libk4FWCorePlugins.so
WARNING duplicated property name 'Input', see https://its.cern.ch/jira/browse/GAUDI-1023
WARNING duplicated property name 'Output', see https://its.cern.ch/jira/browse/GAUDI-1023

Is that OK?

Thanks, Giovanni

m-fila:
Yes, nothing to worry about. It's known warning and expected. From [#213](https://github.com/key4hep/k4FWCore/pull/213#issue-2417169846):
> - Use Input and Output for IOSvc, deprecate input and output since all the other properties are capitalized. The same for IOType instead of ioType. This will print a warning about a duplicated property that should be ignored for now.

giovannimarchiori:
Thanks! Closing the issue then

jmcarcell:
This warning will stop appearing after https://github.com/key4hep/k4FWCore/pull/297 since Gaudi checks only the lowercase name of its properties to emit it.

PR #297, opened 2025-03-17
Remove the deprecated input and output properties in favor of Input and Output

jmcarcell:
Now substituted by their upper-cased variant `Input` and `Output`.

BEGINRELEASENOTES
- Remove the deprecated `input` and `output` properties for IOSvc

ENDRELEASENOTES

This will remove the warning that was reported in https://github.com/key4hep/k4FWCore/issues/240.

### DIFF
--- a/python/k4FWCore/ApplicationMgr.py
+++ b/python/k4FWCore/ApplicationMgr.py
@@ -62,9 +62,7 @@ class ApplicationMgr:
         """
         # First we determine whether we need to peek at all
         inp = None
-        if iosvc_props["input"][0]:
-            inp = "input"
-        elif iosvc_props["Input"][0]:
+        if iosvc_props["Input"][0]:
             inp = "Input"
 
         if not inp:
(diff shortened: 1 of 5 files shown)


python/k4FWCore/ApplicationMgr.py
```diff
@@ -127,17 +125,13 @@ def fix_properties(self):
...
             add_reader = True
         # It seems for a single string the default without a value is '<no value>'
         # while for a list it's an empty list
         # Remove "output" when the corresponding property is removed from IOSvc
-        if (
-            writer is None
-            and (props["output"][0] and props["output"][0] != "<no value>")
-            or (props["Output"][0] and props["Output"][0] != "<no value>")
-        ):
+        if writer is None and (props["Output"][0] and props["Output"][0] != "<no value>"):
```
jmcarcell, line 134:
```suggestion
        if writer is None and props["Output"][0] and props["Output"][0] != "<no value>":
```

PR #298, opened 2025-03-25
Fix reading output from old algorithms when using std::vector in functional algorithms

jmcarcell:
(for an arbitrary number of input collections).

Currently it will fail with an exception if an old algorithm produces a collection and this is read using `std::vector` in the input of an algorithm.

BEGINRELEASENOTES
- Fix reading output from old algorithms when using std::vector in functional algorithms
- Reorganize the tests to make the test `FunctionalMix` read with `std::vector` output from an old algorithm (fails without this fix).

ENDRELEASENOTES


k4FWCore/include/k4FWCore/FunctionalUtils.h
```diff
@@ -142,16 +142,50 @@ namespace details {
...
+                  thisClass->name(),
+                  fmt::format("Failed to cast collection {} to the required type {}, the type of the collection is {}",
+                              handle.objKey(), typeid(EDM4hepType).name(), in ? in->getTypeName() : "[undetermined]"),
+                  StatusCode::FAILURE);
+            }
+          } catch (GaudiException& e) {
+            // When the type of the collection is different from the one requested, this can happen because
+            // 1. a mistake was made in the input types of a functional algorithm
+            // 2. the data was produced using the old DataHandle, which is never going to be in the input type
+            if (e.message().find("different from") != std::string::npos) {
```
tmadlener, line 161: Is it possible to avoid using an exception and message checking here for control flow?

jmcarcell: I guess one could change how stuff is stored by the PodioDataSvc but I'm not sure it's worth it. I don't think it takes that much time since this can happen often but not that often (once per event per collection produced by an old algorithm and per algorithm consuming that collection).

PR #304, opened 2025-04-11
Fix running algorithms with RNTuples as their input and add tests

jmcarcell:
BEGINRELEASENOTES
- Fix running algorithms with RNTuples as their input and add tests, not working before because the TTRee reader in podio was being used
- Add two tests to create an rntuple and read it
- Bump the required version of podio to 1.3 since the function `get_reader` in podio only has support for lists of files as an input after https://github.com/AIDASoft/podio/pull/729

ENDRELEASENOTES


jmcarcell:
I have changed the requirement to podio 1.2.99 until there is a next tag, so this is ready to be merged, unless there are any comments.

PR #305, opened 2025-04-11
Remove the hardcoded collections in the CollectionMerger and fix merging link collections

jmcarcell:
BEGINRELEASENOTES
- Remove the hardcoded collections in the CollectionMerger, using https://github.com/AIDASoft/podio/pull/761. This applies to all collections that are available in EDM4hep.
- Fix link collections that are currently not working since the data type name is not `edm4hep::...LinkCollection` but `podio::LinkCollection...`
- Require a newer version of podio in the CMakeLists.txt 

ENDRELEASENOTES

There is not a tag in podio with https://github.com/AIDASoft/podio/pull/761 so this PR can't be merged yet.

### DIFF
--- a/k4FWCore/components/CollectionMerger.cpp
+++ b/k4FWCore/components/CollectionMerger.cpp
@@ -101,13 +101,9 @@ private:
     const auto ptr = static_cast<T*>(ret);
     const auto sourceColl = static_cast<const T*>(source);
     if (m_copy) {
-      for (const auto& elem : *sourceColl) {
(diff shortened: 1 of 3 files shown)


jmcarcell:
I have changed the requirement to podio 1.2.99 until there is a next tag, so this is ready to be merged, unless there are any comments.

jmcarcell:
Since `::typeName` is not going to be removed (see https://github.com/AIDASoft/podio/pull/775), this can go in. I'll merge today if there aren't any comments.

PR #306, opened 2025-04-24
Make the log-level configurable via command line

tmadlener:
BEGINRELEASENOTES
- Add a `--log-level` flag to `k4run` to make it possible to configure the log level at runtime via the command line
  - accpeted values are `verbose`, `debug`, `info`, `warning` and `error`
  - The log level will apply to the python logging facilities and will also be set in the `ApplicationMgr.OutputLevel`
- Make the python logging a bit more informative by adding the log level as well as some module and function information to the message
- Deprecate the `--verbose` option in favor of `--log-level=verbose`

ENDRELEASENOTES

The main purpose of this (apart from CLI configurability) is the possibility to have some logging facilities for users that go the the same logger if they simply do `logger = k4FWCore.utils.get_logger()` in their scripts.

### DIFF
--- a/k4FWCore/scripts/k4run
+++ b/k4FWCore/scripts/k4run
@@ -175,7 +175,8 @@ def main():
         "--log-level",
         help="Set the log (output) level for python and the ApplicationMgr",
         default="INFO",
-        choices=LOG_LEVELS + tuple(ll.lower() for ll in LOG_LEVELS),
+        type=lambda x: x.upper(),
+        choices=LOG_LEVELS,
     )
 
     # Once to parse the files and another time below to have the new parameters


k4FWCore/python/k4FWCore/utils.py
```diff
@@ -83,3 +85,43 @@ def load_file(opt_file: Union[TextIOWrapper, str, os.PathLike]) -> None:
...
+logging.addLevelName("VERBOSE", logging.VERBOSE)
+
+LOG_LEVELS = ("VERBOSE", "DEBUG", "INFO", "WARNING", "ERROR")
+
+
+def set_log_level(level: str):
+    """Set the passed log level to the k4run logger"""
+    logger = get_logger()
+    if level not in LOG_LEVELS:
+        raise ValueError(f"{level} is not a valid log level name")
```
jmcarcell, line 111:
```suggestion
    if level not in LOG_LEVELS:
        raise ValueError(f"{level} is not a valid log level name")
    logger = get_logger()
```

k4FWCore/scripts/k4run
```diff
@@ -111,27 +111,33 @@ def add_arguments(parser, app_mgr):
     return option_db
 
 
-class LoggingHandler(logging.StreamHandler):
-    def handleError(self, record):
-        # Re-raise SIGPIPE generated exception
-        excp = sys.exc_info()[1]  # from Python 3.11 can be replaced with sys.exception()
-        if isinstance(excp, BrokenPipeError):
-            raise excp
+def parse_log_level(level_str):
```
jmcarcell, line 114: Maybe this function is not needed since `set_log_level` checks and is called early on so if the level is wrong then it's going to fail quickly anyway.

python/k4FWCore/ApplicationMgr.py
```diff
@@ -39,6 +41,12 @@ class ApplicationMgr:
 
     def __init__(self, **kwargs):
         self._mgr = AppMgr(**kwargs)
+        try:
+            self._mgr.OutputLevel = getattr(Configuration, logging.getLevelName(logger.level))
+        except AttributeError:
```
jmcarcell, line 46: This except should never be reached if using `set_log_level` (under normal running). I guess we can leave it here in case someone gets the logger and changes the level manually.

k4FWCore/python/k4FWCore/utils.py
```diff
@@ -83,3 +85,43 @@ def load_file(opt_file: Union[TextIOWrapper, str, os.PathLike]) -> None:
...
 
     exec(code, globals())
+
+
+class LoggingHandler(logging.StreamHandler):
+    def handleError(self, record):
+        # Re-raise SIGPIPE generated exception
+        excp = sys.exc_info()[1]  # from Python 3.11 can be replaced with sys.exception()
+        if isinstance(excp, BrokenPipeError):
+            raise excp
```
jmcarcell, line 95:
```suggestion
        # from Python 3.11 can be replaced with sys.exception()
        if isinstance(excp := sys.exc_info()[1], BrokenPipeError):
            raise excp
```

tmadlener: How far down do we have to go with the python support here? Maybe since we are touching this in any case we could also address the python 3.11 comment? Or are we not there yet?

k4FWCore/scripts/k4run
tmadlener: No clear preference from my side, I think this was there before I realized I had to handle `VERBOSE` as well. The advantage with the current version is that it is very clear for users where the error comes from. But maybe moving the second part of the error message to the `set_log_level` gives them the same information.

python/k4FWCore/ApplicationMgr.py
tmadlener: I would prefer to keep it, because it's not required for users to go through `set_log_level` and this at least gives them a warning that something didn't work.

Should we fall back to `INFO` in that case? Or is that the default already?

k4FWCore/python/k4FWCore/utils.py
```diff
@@ -83,3 +85,43 @@ def load_file(opt_file: Union[TextIOWrapper, str, os.PathLike]) -> None:
...
+
+def get_logger() -> logging.Logger:
+    """Get the logger also used by k4run"""
+    global _logger
+    if _logger is not None:
+        return _logger
+    _logger = logging.getLogger()
+    _logger.setLevel(logging.INFO)
+    formatter = logging.Formatter("[k4run - %(levelname)s] %(module)s.%(funcName)s: %(message)s")
+    handler = logging.StreamHandler(sys.stdout)
```
jmcarcell, line 123:
```suggestion
    handler = LoggingHandler(sys.stdout)
```
I'm realizing now that `LoggingHandler` is not used anywhere and indeed piping doesn't work well in all cases:
```
k4run --dry-run ExampleFunctionalProducer.py | head -n 1
```
gives me a bunch of errors like:
```
[k4run] --> ExampleFunctionalProducer
--- Logging error ---
Traceback (most recent call last):
  File "/usr/lib/python3.13/logging/__init__.py", line 1154, in emit
    self.flush()
    ~~~~~~~~~~^^
  File "/usr/lib/python3.13/logging/__init__.py", line 1136, in flush
    self.stream.flush()
    ~~~~~~~~~~~~~~~~~^^
BrokenPipeError: [Errno 32] Broken pipe
Call stack:
  File "/k4FWCore/install/bin/k4run", line 258, in <module>
    main()
  File "/k4FWCore/install/bin/k4run", line 213, in main
    logger.info(f"Option name: {optionName} {opts_dict[optionName]}")
Message: 'Option name: IOSvc.CollectionNames []'
```

k4FWCore/scripts/k4run
jmcarcell: I think I would remove, one less function to maintain.

python/k4FWCore/ApplicationMgr.py
jmcarcell: If using k4run then `INFO` is the default, if not I'm not sure.

k4FWCore/scripts/k4run
tmadlener: I always forget that `add_argument` takes a `choices` kwarg, so there is no need for custom handling. I have switched to that and added more information to the error message in `set_log_level`.

k4FWCore/scripts/k4run
```diff
@@ -111,27 +111,22 @@ def add_arguments(parser, app_mgr):
...
-            raise excp
+class DeprecatedStoreTrueAction(argparse.Action):
+    def __init__(self, option_strings, dest, nargs=0, **kwargs):
+        super().__init__(option_strings, dest, nargs=nargs, **kwargs)
 
-        super().handleError(record)
+    def __call__(self, parser, namespace, values, option_string=None):
+        warnings.warn(
+            f"Argument '{self.dest}' is deprecated and will be removed in a future version.",
+            DeprecationWarning,
```
tmadlener, line 121:
```suggestion
            FutureWarning,
```
Otherwise it will not show up by default.

k4FWCore/python/k4FWCore/utils.py
jmcarcell: Fixed in https://github.com/key4hep/k4FWCore/pull/313

k4FWCore/python/k4FWCore/utils.py
jmcarcell: Outdated since https://github.com/key4hep/k4FWCore/pull/313

k4FWCore/scripts/k4run
tmadlener: This seems to have been lost in the rebasing / merge conflict resolution (i.e. the removal of the `parse_log_level`)

jmcarcell: Argh that was probably that the branch I had wasn't up to date with the remote, and since I saw most of the stuff I assumed it was. Do you still have the commit if you didn't pull?

jmcarcell: I'll just try it in a commit since it should be an easy change

tmadlener: I can also try to do the rebase push back that commit. I still have it locally

jmcarcell: Ok, then I don't do anything :smile:

tmadlener: :+1: Done.

k4FWCore/scripts/k4run
```diff
@@ -153,11 +171,21 @@ def main():
         help="Run with ROOT in interactive mode (e.g. to see plots)",
         default=False,
     )
+    parser.add_argument(
+        "--log-level",
+        help="Set the log (output) level for python and the ApplicationMgr",
+        default="INFO",
+        choices=LOG_LEVELS + tuple(ll.lower() for ll in LOG_LEVELS),
```
jmcarcell, line 178:
```suggestion
        type=lambda x: x.upper(),
        choices=LOG_LEVELS,
```
This is an alternative, but it doesn't print all the choices in the help. I think the help becomes a bit ugly with every choice duplicated:
```
usage: k4run [--dry-run] [-v] [-n NUM_EVENTS] [-l] [--gdb] [--interactive-root]
             [--log-level {VERBOSE,DEBUG,INFO,WARNING,ERROR,verbose,debug,info,warning,error}]
```
and also it's not even aligned like most options:
```
  --gdb                 Attach gdb debugger
  --interactive-root    Run with ROOT in interactive mode (e.g. to see plots)
  --log-level {VERBOSE,DEBUG,INFO,WARNING,ERROR,verbose,debug,info,warning,error}
                        Set the log (output) level for python and the ApplicationMgr
  -h, --help            show this help message and exit
```

PR #307, opened 2025-04-25
Clean up several CMakeLists.txt files and move python files to a single place

jmcarcell:
to `k4FWCore/python`. Two custom commands can be removed since what they do is already done by another custom command when building the tests.

BEGINRELEASENOTES
- Clean up several CMakeLists.txt files. For example, remove redundant calls to `find_package`.
- Move python files that will be installed to a single place

ENDRELEASENOTES

Tests can be run also without installing (I think this was also possible before but I checked since the CI does install before running).


tmadlener:
> I think this was also possible before but I checked since the CI does install before running

I can confirm that even without this PR it's possible to run the tests without installing first.

jmcarcell:
If there aren't any comments I'll merge this tomorrow, nothing should change with this cleanup.

Issue #229, opened 2024-08-29
Option to print timing info?


alexandertuna:
Hi developers, @Lrozzy and I have a small question: is there an option to print timing information with k4run, either at the CL or in the algs? e.g. something like:

```
[MESSAGE "Marlin"]  Total: 2.096301e+02 s in 3 events  ==> 6.987670e+01 [ s/evt.]
```

I thought this feature existed, but I can't find it, so perhaps I imagined it. Do you know if this feature exists?

Thanks!
Alex

jmcarcell:
Try this, it seems to be accurate enough from my testing:
``` python
from Configurables import AuditorSvc, AlgTimingAuditor
auditor = AuditorSvc()
auditor.Auditors = [AlgTimingAuditor()]
```
and then you add `auditor` to the list of external services passed to `ApplicationMgr` like:
``` python
ApplicationMgr(
    TopAlg=[producer],
    EvtSel="NONE",
    EvtMax=10 if not args[0].second else 20,
    ExtSvc=[EventDataSvc("EventDataSvc"), auditor],
    OutputLevel=INFO,
)
```

m-fila:
`AlgTimingAuditor` should be the way to go

But in case you'd like to later record some more information then there is also `TimelineSvc` that can output simple csv file with start and stop (ns) timestamps for each algorithm

```python
from Configurables import TimelineSvc
TimelineSvc(RecordTimeline=True, DumpTimeline=True, TimelineFile="timeline.csv" )
```

alexandertuna:
Thanks very much @jmcarcell and @m-fila ! Both of these suggestions look great. We'll use them.

I'm happy to close the issue now but please feel free to reopen in case something else comes to mind.

PR #308, opened 2025-04-28
Add documentation about how to instrument Gaudi

tmadlener:
BEGINRELEASENOTES
- Add some documentation about the `AlgTimingAuditor` and the `TimelineSvc` to show how timing information can be obtained from Gaudi at different levels of details.

ENDRELEASENOTES

- See #229 and key4hep/k4MarlinWrapper#201

### DIFF
--- a/doc/algorithmTiming.md
+++ b/doc/algorithmTiming.md
@@ -33,7 +33,7 @@ the following
 
 ```python
 from Configurables import AuditorSvc, AlgTimingAuditor, EventDataSvc
-from k4FWCore import AppliationMgr
+from k4FWCore import ApplicationMgr
 
 # List of your algorithms that you want to run
 algList = []
--- a/doc/algorithmTiming.md
+++ b/doc/algorithmTiming.md
@@ -148,3 +148,11 @@ each line has
 The event number is a sequential number inside Gaudi and does not necessarily
 correspond with the event number that can be obtained from the `EventHeader`
 ```
+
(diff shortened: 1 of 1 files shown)


m-fila:
The csv output from timeline service can be visualized with `hivetimeline.py` (should be in path together with other gaudi utilities) although its not very beautiful
timeline

doc/algorithmTiming.md
```diff
@@ -0,0 +1,158 @@
...
+## `AlgTimingAuditor`
+
+The `AlgTimingAuditor` in combination with the `AuditorSvc` allow to easily get
+an overview over the execution time of a workflow and in which algorithms,
+services or tools time is spent. In order to configure it, you need to simply do
+the following
+
+```python
+from Configurables import AuditorSvc, AlgTimingAuditor, EventDataSvc
+from k4FWCore import AppliationMgr
```
jmcarcell, line 36:
```suggestion
from k4FWCore import ApplicationMgr
```

jmcarcell (approved):
Looks good. I find using `perf` easier most of the time since the steering file doesn't need to be changed to see what's taking long.

tmadlener:
Yeah, for serious profiling `perf` is most likely the way to go, but that might require permissions that you usually don't have on e.g. lxplus or other institute managed machines. But for some first numbers I think this is good enough and we don't want to throw people into the `perf` rabbit hole immediately ;).

PR #310, opened 2025-04-29
Populate correct meta information in load_file, deprecate taking file objects

tmadlener:
BEGINRELEASENOTES
- Make sure that `load_file` sets the correct `__file__` and `__spec__` information for files that are loaded with it
- Restore the output of the filename in case loading doesn't work
- Deprecate passing a file handle to `load_file` and always pass a path-like object

ENDRELEASENOTES

Setting the correct `__file__` is necessary as that is quite frequently used for some special casing, or also to add things to `sys.path` relative to the path of the loaded file, e.g. [here](https://github.com/iLCSoft/ILDConfig/blob/d522d2ded16031bcce9e6a042c75fa1abb1bef3e/StandardConfig/production/ILDReconstruction.py#L22-L24)

Putting out the correct file name was lost in https://github.com/key4hep/k4FWCore/pull/213

Finally it looks like a simple change in `k4run` makes it trivial to pass in path-like objects rather than open file handles (which we didn't bother to close before).

### DIFF
--- a/k4FWCore/python/k4FWCore/utils.py
+++ b/k4FWCore/python/k4FWCore/utils.py
@@ -25,6 +25,8 @@ from importlib.machinery import SourceFileLoader
 import importlib.util
 from pathlib import Path
 
+import warnings
+
 
 def check_wrong_imports(code: str) -> None:
     """Check for wrong imports in the given code.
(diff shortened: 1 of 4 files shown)


python/k4FWCore/utils.py
```diff
@@ -77,14 +80,39 @@ def load_file(opt_file: Union[TextIOWrapper, str, os.PathLike]) -> None:
...
+            {
+                "__file__": os.path.realpath(opt_file),
+                "__spec__": importlib.util.spec_from_loader(loader.name, loader),
+            }
+        )
     else:
+        warnings.warn(
+            "load_file will remove support for handling TextIOWrapper. Please switch to pasing os.PathLike",
+            FutureWarning,
+        )
```
tmadlener, line 108: I am not sure this is used anywhere. It might be possible to remove this entirely and clean the implementation up a bit more.

PR #312, opened 2025-05-12
[Collid assignment

tmadlener:
BEGINRELEASENOTES
- Thank you for writing the text to appear in the release notes. It will show up
  exactly as it appears between the two bold lines
- ...

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/include/k4FWCore/FunctionalUtils.h
+++ b/k4FWCore/include/k4FWCore/FunctionalUtils.h
@@ -29,6 +29,7 @@
 #include <GaudiKernel/GaudiException.h>
 
 #include "podio/CollectionBase.h"
+#include <podio/CollectionIDTable.h>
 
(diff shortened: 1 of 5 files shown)


k4FWCore/include/k4FWCore/ICollectionFromObjectSvc.h
```diff
@@ -0,0 +1,43 @@
...
+#include <podio/ObjectID.h>
+#include <podio/utilities/TypeHelpers.h>
+
+#include <fmt/format.h>
+
+class ICollectionFromObjectSvc : virtual public IInterface {
+public:
+  DeclareInterfaceID(ICollectionFromObjectSvc, 1, 0);
+
+  template <podio::ObjectType O>
```
tmadlener, line 34: This could also be changed to something that works with podio 1.3, as this is not part of that tag.

PR #313, opened 2025-05-14
Fix using k4run with pipes, like `| head`, and add a test

jmcarcell:
BEGINRELEASENOTES
- Use signal to stop immediately after receiving a SIGPIPE. Before, the output from k4run itself would always print and the pipe command (for example `| head`) would only be applied to the output from Gaudi.
- Add a test that uses pipes

ENDRELEASENOTES

Found in https://github.com/key4hep/k4FWCore/pull/306#discussion_r2066168920. See also https://github.com/key4hep/k4FWCore/pull/186, when handling of SIGPIPE was initially introduced.

### DIFF
--- a/test/k4FWCoreTest/CMakeLists.txt
+++ b/test/k4FWCoreTest/CMakeLists.txt
@@ -223,9 +223,7 @@ add_test_with_env(TypeMisMatchDemoMultiple options/TypeMisMatchDemoMultiple.py P
 add_test_with_env(ReadMarlinWrapperCollection options/readMarlinWrapperCollection.py)
 
 add_test(NAME check_broken_pipe
-         COMMAND bash -c "${K4RUN} options/ExampleFunctionalProducer.py | head -n 2 | wc -l"
+         COMMAND bash -c "[ $(${K4RUN} options/ExampleFunctionalProducer.py | head -n 2 | wc -l) = 2 ]"
          WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
        )
-
-set_tests_properties(check_broken_pipe PROPERTIES PASS_REGULAR_EXPRESSION "^2")
 set_test_env(check_broken_pipe)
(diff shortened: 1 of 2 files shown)


test/k4FWCoreTest/CMakeLists.txt
```diff
@@ -221,3 +221,11 @@ add_test_with_env(TypeMisMatchDemo options/TypeMisMatchDemo.py PROPERTIES PASS_R
 
 add_test_with_env(TypeMisMatchDemoMultiple options/TypeMisMatchDemoMultiple.py PROPERTIES PASS_REGULAR_EXPRESSION "Failed to cast collection MCParticles1 to the required type")
 add_test_with_env(ReadMarlinWrapperCollection options/readMarlinWrapperCollection.py)
+
+add_test(NAME check_broken_pipe
+         COMMAND bash -c "${K4RUN} options/ExampleFunctionalProducer.py | head -n 2 | wc -l"
+         WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
+       )
+
+set_tests_properties(check_broken_pipe PROPERTIES PASS_REGULAR_EXPRESSION "^2")
```
jmcarcell, line 230:
```suggestion
set_tests_properties(check_broken_pipe PROPERTIES PASS_REGULAR_EXPRESSION "^2")
```
This could be more strict but all variants of `"^2$"` that I tried didn't work :shrug:

test/k4FWCoreTest/CMakeLists.txt
```diff
@@ -221,3 +221,11 @@ add_test_with_env(TypeMisMatchDemo options/TypeMisMatchDemo.py PROPERTIES PASS_R
 
 add_test_with_env(TypeMisMatchDemoMultiple options/TypeMisMatchDemoMultiple.py PROPERTIES PASS_REGULAR_EXPRESSION "Failed to cast collection MCParticles1 to the required type")
 add_test_with_env(ReadMarlinWrapperCollection options/readMarlinWrapperCollection.py)
+
+add_test(NAME check_broken_pipe
+         COMMAND bash -c "${K4RUN} options/ExampleFunctionalProducer.py | head -n 2 | wc -l"
```
tmadlener, line 226:
```suggestion
         COMMAND bash -c "[ $(${K4RUN} options/ExampleFunctionalProducer.py | head -n 2 | wc -l) = 2 ]"
```
Would make the it possible to go without the regular expression checking below

PR #316, opened 2025-05-16
Use the Warnings property only if it exists for EventLoopMgr

jmcarcell:
It has been removed in Gaudi in https://gitlab.cern.ch/gaudi/Gaudi/-/merge_requests/993, which will appear first in 40.0. This is currently silencing these two warnings:

```
  EventLoopMgr      WARNING Unable to locate service "EventSelector"
  EventLoopMgr      WARNING No events will be processed from external input.
```

that would always appear otherwise since we don't use the EventSelector. These changes change the output level of the default EventLoopMgr to WARNING but since these two warnings have been demoted to INFO, they will still be silenced.

BEGINRELEASENOTES
- Use the Warnings property only if it exists for EventLoopMgr
- Do not use the Warnings property in the tests since it is going to be removed (warnings will not be silenced in these tests before v40r0)

ENDRELEASENOTES
https://gitlab.cern.ch/gaudi/Gaudi/-/merge_requests/1740 has been merged and it will appear in v40, giving us a way of knowing if we are on 40.0.

### DIFF
--- a/python/k4FWCore/ApplicationMgr.py
+++ b/python/k4FWCore/ApplicationMgr.py
@@ -114,12 +114,10 @@ class ApplicationMgr:
         if not hasattr(self._mgr, "EventLoop"):
             self._mgr.EventLoop = EventLoopMgr()
             try:
-                # Only available in Gaudi >= 40.0.0, when the Warnings attribute is removed
-                from Gaudi import __version__  # noqa: F401
-
-                self._mgr.EventLoop.OutputLevel = WARNING
-            except ImportError:
                 self._mgr.EventLoop.Warnings = False
+            # Warnings doesn't exist after v40
+            except AttributeError:
+                self._mgr.EventLoop.OutputLevel = WARNING
 
         if "MetadataSvc" in self._mgr.allConfigurables:
             self._mgr.ExtSvc.append(self._mgr.allConfigurables["MetadataSvc"])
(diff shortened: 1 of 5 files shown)


jmcarcell:
I'll merge this later today if there are no more comments. It works in CI (before Gaudi v40) and locally for me (Gaudi master) I need it, otherwise there are errors about Warnings not existing when running k4run.

python/k4FWCore/ApplicationMgr.py
```diff
@@ -110,10 +111,15 @@ def _setup_reader(self, reader, iosvc_props):
...
         # This will suppress two warnings about not using external input
-        try:
-            self._mgr.EventLoop
-        except AttributeError:
-            self._mgr.EventLoop = EventLoopMgr(Warnings=False)
+        if not hasattr(self._mgr, "EventLoop"):
+            self._mgr.EventLoop = EventLoopMgr()
+            try:
+                # Only available in Gaudi >= 40.0.0, when the Warnings attribute is removed
+                from Gaudi import __version__  # noqa: F401
```
tmadlener, line 118: Does this mean that `Gaudi.__version__` is only available from Gaudi 40.0.0 onwards? Does `self._mgr.EventLoop.OutputLevel` not raise an `AttributeError` that we could catch instead of relying on this import?

tmadlener (approved):

jmcarcell:
And today CI is fine while yesterday it was impossible to make it pass, it's a very weird error the one about the `[0]` indexing (and then it complains with `module` not callable)

Issue #292, opened 2025-02-25
Deprecate: PodioInput and k4DataSvc


andresailer:
Only the IOSvc must be used for new code from now on

- [ ] Add deprecation warnings
- [ ] Document migration towards IOSvc
- [ ] Eventually remove PodioInput / k4DataSvc

m-fila:
The services to be deprecated are still mentioned in the documentation:
- key4hep-doc  - https://github.com/key4hep/key4hep-doc/pull/92
- key4hep-tutorials
- k4MarlinWrapper

tmadlener:
We need to make sure that `Gaudi::Algorithm`s are still supported, I think. They still use `DataHandle` and there might be some subtleties in the fact that a `DataHandle` doesn't store things as an `AnyDataWrapper` into the TES. This is currently causing an issue in https://github.com/key4hep/k4MarlinWrapper/issues/219

tmadlener:
These are also injected by default in k4MarlinWrapper: https://github.com/key4hep/k4MarlinWrapper/issues/231

tmadlener:
Looking at a bit more code and also the `DataHandle`, I am wondering whether it's possible to transparently switch to the way we interact with the TES by replacing the implementation of the `DataHandle` to use that. IIUC, than the main thing that needs to happen for our Functional utils to pick things up correctly collections need to go into the `TES` as an `AnyDataWrapper<std::unique_ptr<podio::CollectionBase>>`, right? In that case, I think it should be possible to implement the `DataHandle` on top of that.

Is that an avenue we want to explore? Or is there a better way (aka more "Gaudi native") way to do this?

PR #318, opened 2025-05-19
[WIP] Deprecate the PodioDataSvc and friends

tmadlener:
BEGINRELEASENOTES
- Deprecate `PodioDataSvc` (and `k4DataSvc` and `FCCDataSvc`) as well as the `PodioInput`, `PodioOutput`.
- Add `WARNING` messages at runtime to inform users about the upcoming removal.

ENDRELEASENOTES

- [ ] Includes #317 

This is the first and easy part of deprecating the `PodioDataSvc` (see #292). Moving to non-deprecated versions wherever possible internally still needs to be done. I think it should be possible to implement the `k4FWCore::DataHandle` without the need of a custom `DataWrapper` that relies on the `PodioDataSvc`. The idea is to place things going through the `k4FWCore::DataHandle` into the TES the same way as the `IOSvc` / Functional utils do. That way Functional and non Functional algorithms should integrate seamlessly.


m-fila:
Could we also deprecate `MetaDataHandle` in favor of `MetadataSvc` (or `k4FWCore::getParameter/putParameter` which are resolving the services internally)?

tmadlener:
Up for discussion IMO. Technically the `MetaDataHandle` and the `MetadataSvc`/`MetadataUtils` serve slightly different purposes at the moment (at least partially). The `MetaDataHandle` has the constructor taking a `DataHandle` as well, so it allows to easily set "collection parameters" (which are really just parameters with a different naming convention), while for the `k4FWCore::{get,put}Parameter` you have to be actively aware of that naming convention, i.e. `podio::collMetadataParamName`)

m-fila:
Then what is the correct approach to use collection metadata in functional algorithms? They don't have any data-handles there that could be passed to metadata-handles

m-fila:
During meeting we discussed that collection metadata in functional algorithms is a separate issue. We won't deprecate the metadata-handles here.

tmadlener:
Strategic question: How do we want to do the phaseout here? Until we fully remove the `PodioDataSvc`, etc, builds of k4FWCore will have deprecation warnings pop up all over the place. On the other hand I don't think we can remove it in this PR as that would mean that there is no chance for people working on tagged releases to realize that this is even deprecated and will be removed.

Some of the warnings in k4FWCore builds could probably be fixed with some pre-processor toggling of `-Wdeprecated-declaration`, but I am not sure it's possible for all, nor whether we actually want to do that.

tmadlener:
I am not entirely sure why the key4hep CI workflows are not passing. In principle we already pass `-Wno-error=deprecated-declarations` via CMake: https://github.com/key4hep/key4hep-actions/blob/706b9a775f4d22a2603ed36ece4196e972038087/key4hep-build/action.yml#L123

However, that doesn't seem to arrive at the compiler invocation at the end: https://github.com/key4hep/k4FWCore/actions/runs/16002321535/job/45140217866?pr=318#step:4:266 (only `-Werror` appears here).

I cannot reproduce this locally even when invoking CMake the same way as we do in CI...

andresailer:
I think the quotes get lost and the later compiler flags get lost
```
cmake .. -DCMAKE_CXX_STANDARD=20 -DCMAKE_INSTALL_PREFIX=../install -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_FLAGS=-Werror -Wno-error=cpp -Wno-error=deprecated-declarations -G Ninja
```
They are not in the compiler invocation.

tmadlener:
Are we still waiting on something for this one? I think it would be nice to have these messages out now so that we have some time for people to adapt before we remove things.

jmcarcell:
Good for me, but people using the Marlin wrapper still are not using IOSvc so this warning will be there even though the alternative is not there yet?

PR #322, opened 2025-05-26
Add the EfficiencyFilter algorithm and tests running it

jmcarcell:
BEGINRELEASENOTES
- Add an EfficiencyFilter, an algorithm that will take an EDM4hep collection and return a subset collection with a certain efficiency. The `Exact` parameter can be used to control whether the efficiency is exactly that number of simply
  each element is kept with `Efficiency` probability.
- Add a test that runs the EfficiencyFilter both in the exact version (to check that exactly `Efficiency * length` elements are saved) and in the non-exact version

ENDRELEASENOTES
  
It's also an example of a transformer that works on a single arbitrary collection, of which we didn't have any examples before.

See https://github.com/key4hep/k4Reco/pull/27#issuecomment-2908670868

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Algorithm merging multiple collections of the same type into a single collection
 
 ### EfficiencyFilter
 
-Algorithm that takes a collection and creates a subset collection with a random sample of the elements of the collection.
+Algorithm that takes a collection and creates a subset collection by randomly accepting elements according to the user-defined `Efficiency`.
 
 ### EventHeaderCreator
 


k4FWCore/include/k4FWCore/FunctionalUtils.h
```diff
@@ -67,6 +67,13 @@ namespace details {
     return static_cast<const T&>(*arg);
   }
 
+  // This is used by the EfficiencyFilter
+  template <typename T, typename P>
+    requires std::same_as<P, podio::CollectionBase* const>
+  const auto& maybeTransformToEDM4hep(P& arg) {
+    return static_cast<const podio::CollectionBase&>(*arg);
+  }
+
```
m-fila, line 76: I think this is some artifact that isn't used now

k4FWCore/components/EfficiencyFilter.cpp
```diff
@@ -0,0 +1,138 @@
...
+    const auto ptr = static_cast<T*>(ret);
+    const auto sourceColl = static_cast<const T*>(source);
+
+    if (m_exact) {
+      // If exact, we will keep exactly the number of elements given by the efficiency
+      const size_t nToKeep = static_cast<size_t>(static_cast<double>(sourceColl->size()) * m_efficiency);
+
+      std::vector<size_t> indexes(sourceColl->size());
+      std::iota(indexes.begin(), indexes.end(), 0);
+      std::ranges::shuffle(indexes, randomGen);
```
andresailer, line 123: Take the first `nToKeep` indices and sort them, so that the order of hits is the same as in the original.

k4FWCore/components/EfficiencyFilter.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include <random>
+#include <string>
+
+struct EfficiencyFilter final : k4FWCore::Transformer<podio::CollectionBase*(const podio::CollectionBase&,
+                                                                             const edm4hep::EventHeaderCollection&)> {
+  EfficiencyFilter(const std::string& name, ISvcLocator* svcLoc)
+      : Transformer(name, svcLoc, {KeyValues("InputCollection", {""}), KeyValues("EventHeader", {"EventHeader"})},
+                    {KeyValues("OutputCollection", {""})}) {
+
+    if (System::cmdLineArgs()[0].find("genconf") != std::string::npos) {
```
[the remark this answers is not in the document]
jmcarcell: Removed

k4FWCore/components/EfficiencyFilter.cpp
jmcarcell: This is now done

k4FWCore/components/EfficiencyFilter.cpp
```diff
@@ -0,0 +1,137 @@
...
+               ? (result = createSubsetCollection<typename T::collection_type>(&coll, uid))
+               : nullptr));
+    return result;
+  }
+
+  template <typename T>
+  podio::CollectionBase* createSubsetCollection(const podio::CollectionBase* source, const size_t uid) const {
+
+    std::mt19937 randomGen(std::random_device{}());
+    randomGen.seed(uid);
```
m-fila, line 105:
```suggestion
    std::mt19937 randomGen(uid);
```

Seed can be set directly in the constructor

m-fila:
This won't work with the data-model extensions, right?

k4FWCore/components/EfficiencyFilter.cpp
```diff
@@ -0,0 +1,136 @@
...
+    debug() << "EfficiencyFilter called for collection of type " << coll.getTypeName() << endmsg;
+
+    const auto uid = m_uniqueIDSvc->getUniqueID(evtHeader, name());
+
+    auto ptr = dispatchByType(
+        coll, uid,
+        edm4hep::edm4hepDataTypes{}); // Dispatch to the correct createSubsetCollection template function based on type
+    if (!ptr) {
+      ptr = dispatchByType(coll, uid, edm4hep::edm4hepLinkTypes{}); // Try link types if not found in data types
+    }
```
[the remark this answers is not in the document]
jmcarcell: This is now done in the algorithm, and `dispatchByType` is called only once.

jmcarcell:
> This won't work with the data-model extensions, right?

No, like any algorithm we create here, it can't know about the extensions. It's always the case with any algorithm, no?

BrieucF:
> I am wondering whether we can have a somewhat lightweight pseudo plugin mechanism here that makes it possible to add another type list downstream where extensions are defined. OTOH, not having all the functionality for all possible extensions probably also serves as a good incentive for people to actually go use EDM4hep.

I would argue here that since we have the policy to integrate in EDM4hep new data types only if they have been thoroughly tested, we'd rather provide the needed functionalities to fully integrate extensions into realistic workflows.

jmcarcell:
You should be able to lose less hits in `DDPlanarDigi` with
``` python
alg.ForceHitsOntoSurface = True
alg.MinEnergy = -1
alg.UseTimeWindow = False
alg.MaxTries = 1000
```

The minimum energy in case there are hits with energy 0 (probably not). Then `UseTimeWindow` is false by default and the one that is unavoidable in the algorithm right now is this `continue` after trying to smear the hit several times: https://github.com/key4hep/k4Reco/blob/main/k4Reco/DDPlanarDigi/components/DDPlanarDigi.cpp#L272, so maybe with more tries you get more hits. You can also check the efficiency after you select 80% since depending on how many hits you may not have 80% exactly and 75% could be the closest and 69% just one hit less.

jmcarcell:
@mahmoudali2 had an interesting use-case, where there were lots of events with few hits each. The current default of always selecting a percentage of hits for each event in that case makes the ratio of accepted_hits / total_hits over all events be very different from what is the input. I think that it makes sense that the default is for each hit to be accepted with `Efficiency` probability, then the ratio accepted / total over all events will be preserved independently of the number of hits per event.
If there are any more comments I'll merge this today or tomorrow.

README.md
```diff
@@ -24,6 +24,10 @@ See the [documentation](doc/PodioInputOutput.md) for more information.
 
 Algorithm merging multiple collections of the same type into a single collection.
 
+### EfficiencyFilter
+
+Algorithm that takes a collection and creates a subset collection with a random sample of the elements of the collection.
```
BrieucF, line 29:
```suggestion
Algorithm that takes a collection and creates a subset collection by randomly rejecting elements according to the user defined `Efficiency`.
```

BrieucF: I am not sure there is a use case for `Exact = True` but ok, it is there already

mahmoudali2 (commented):
I tested the last updated  with 0.8 efficiency and found that:
- filter_not_exact with Exact= False gives --> 0.8006 
- filter gives 0.797

test/k4FWCoreTest/options/ExampleEfficiencyFilter.py
```diff
@@ -0,0 +1,94 @@
...
+    "CollectionMerger",
+    InputCollections=["MCParticles1"] * 10,
+    OutputCollection=["MergedMCParticles"],
+)
+
+merger_links = CollectionMerger(
+    "LinksMerger",
+    InputCollections=["Links"] * 10,
+    OutputCollection=["MergedLinks"],
+)
```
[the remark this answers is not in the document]
BrieucF: Hi Mahmoud, I don't think so, it is just here as an example.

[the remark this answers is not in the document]
jmcarcell: Yes, it's only an example, I think in that case is only to have more links

jmcarcell:
> I tested the last updated with 0.8 efficiency and found that:
> 
>     * filter_not_exact with Exact= False gives --> 0.8006
> 
>     * filter gives 0.797

Yes, now `Exact=False` is the default, so you'll get the same setting it or not (they are different numbers because they get different seeds because the algorithm name is different).

> @jmcarcell is this expected to work on extension datatypes?

No, none of the algorithms written without knowledge about the extension will work as it is now. How this one works (and the other generic ones) is whatever the algorithm does is a template that is instantiated for every type in EDM4hep. I'm not sure how it would be possible to write generic algorithms like this one that also work for extensions when `k4FWCore` can't know anything about these extensions. There is a bit ugly workaround, which is to copy the algorithm to where the extension is defined and use there the types provided by the extension in addition to the ones from EDM4hep.

PR #323, opened 2025-05-26
Explicitly mention source files instead of GLOB

tmadlener:
spack doesn't always re-run cmake when in a development environment leading to failing builds due to left over source file mentions (e.g. when switching branches)

BEGINRELEASENOTES
- Switch from `file(GLOB ...)` to explicitly mention source files for k4FWCore plugin sources

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/CMakeLists.txt
+++ b/k4FWCore/CMakeLists.txt
@@ -28,10 +28,22 @@ target_include_directories(k4FWCore PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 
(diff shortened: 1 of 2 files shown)


jmcarcell:
I don't know what is better. The list changes so slowly that this is annoying very few times, however it is convenient not to have to specify what is built (easy to forget), but this is also convenient very rarely. The tests are in the same situation and those change much more: https://github.com/key4hep/k4FWCore/blob/master/test/k4FWCoreTest/CMakeLists.txt#L20
I feel like since we are following the convention of only having components in the `components` folder then we shouldn't need specify what we want to build, it's already there.

jmcarcell:
I was thinking about this and this is probably better. It would make easier to develop when touching the headers that are included by every file by commeting out some files, otherwise everything has to be compiled; I have done this in the past. It may make cmake a few ms faster too. I have specified the files that are going to be compiled for tests so I think this is ready to be merged.

PR #328, opened 2025-06-12
Remove the outdated script k4-print-joboptions

jmcarcell:
BEGINRELEASENOTES
- Remove the outdated script k4-print-joboptions. It doesn't work on current files, it's not using the podio readers (so it doesn't support RNTuples, if it was working) and the same functionality can easily be obtained with `podio-dump`.

ENDRELEASENOTES

Unchanged since early 2021.


tmadlener:
With `podio-dump` you still have to post-process the output though, right? While looking at #329 I was thinking that having a simple tool to retrieve them again would be nice. I wasn't aware we had one even though it no longer works.

As already mentioned in #329 if we store them, I think we should also be able to retrieve them and get a valid (and equivalent to the initial input) configuration back so that the same job could in principle be run again.

PR #329, opened 2025-06-12
Do not set properties twice for ApplicationMgr and MessageSvc in the job options

jmcarcell:
It seems the options service has all these and it looks like it has had them for a long time. I checked and it's the same properties. For the NTupleSvc, since no one is using it I propose to simply remove them.

BEGINRELEASENOTES
- Do not set properties twice for ApplicationMgr and MessageSvc. Do not write any properties for the NTupleSvc
- Construct strings in place instead of using `std::stringstream`

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/components/PodioOutput.cpp
+++ b/k4FWCore/components/PodioOutput.cpp
@@ -115,14 +115,12 @@ StatusCode PodioOutput::finalize() {
   const auto& jobOptionsSvc = Gaudi::svcLocator()->getOptsSvc();
   const auto& configured_properties = jobOptionsSvc.items();
   for (const auto& per_property : configured_properties) {
-    std::stringstream config_stream;
     // sample output:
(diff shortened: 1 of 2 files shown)


jmcarcell:
I can add a test but it's not trivial. First, we don't have a defined format, and the current one is a mix of things. I'm not sure about checking exactly the properties that exist now since this can change for example from Gaudi and we have to fix it.

For the format currently everything goes between quotes but this creates cases where the value is a string that is also in quotes: `b'TimelineSvc.TimelineFile = "\'timeline.csv\''"\n`. We could get rid of the `\n` character after every parameter since they are already different strings, but then podio-dump won't display them in different lines even though when reading from podio one gets a list.

tmadlener:
OK. I think at some point we should add a check to see whether we can run the config that we store.

PR #335, opened 2025-07-25
Get the available collections from IOSvc instead of peeking in the file

jmcarcell:
when running single-threaded to improve the performance since it's not necessary to read the input file before running to find how many events or which collections exist. This was added in the past because it's needed for running multithreaded. For multithreading a fix has been discussed in Gaudi, because there is not a way to tell the event loop to stop right now because there are no more events, but it will take some time.

BEGINRELEASENOTES
- Get the available collections from IOSvc instead of peeking in the file (if any) before running, improving the performance when running single-threaded (unchanged when running multithreaded)

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/components/Reader.cpp
+++ b/k4FWCore/components/Reader.cpp
@@ -40,25 +40,22 @@ class CollectionPusher : public Gaudi::Functional::details::BaseClass_t<Gaudi::F
 
   template <typename T>
   using OutputHandle_t = Gaudi::Functional::details::OutputHandle_t<Traits_, std::remove_pointer_t<T>>;
-  std::vector<OutputHandle_t<Out>> m_outputs;
 
 protected:
+  std::vector<OutputHandle_t<Out>> m_outputs;
+
   Gaudi::Property<std::vector<std::string>> m_inputCollections{
-      this, "InputCollections", {"First collection"}, "List of input collections"};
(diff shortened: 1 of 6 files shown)


k4FWCore/components/Reader.cpp
```diff
@@ -40,27 +40,24 @@ class CollectionPusher : public Gaudi::Functional::details::BaseClass_t<Gaudi::F
...
 protected:
+  std::vector<OutputHandle_t<Out>> m_outputs;
+
   Gaudi::Property<std::vector<std::string>> m_inputCollections{
-      this, "InputCollections", {"First collection"}, "List of input collections"};
-  // Gaudi::Property<std::string>                                        m_input{this, "Input", "Event", "Input file"};
+      this, "InputCollections", {}, "List of input collections"};
 
 public:
   CollectionPusher(std::string name, ISvcLocator* locator)
```
andresailer, line 51: Is there some documentation about what this `CollectionPusher` does?

jmcarcell: It's the base class of the Reader, it could be a single class and I plan on rewriting it. I think this is how it was done in LHCb by the time it was written, now I'm not sure if there is any purpose, certainly we do not have other readers that could use it.

andresailer: Ok, thanks for the explanation!

andresailer (approved):

PR #339, opened 2025-08-06
Reserve when possible and use better names for the variables

jmcarcell:
BEGINRELEASENOTES
- Clean up of functional utilities and code: reserve when possible and use better names for the variables and use `const` in loops when possible
- Fix exceptions to be constructed the same way with `fmt::format`.

ENDRELEASENOTES

Some of the variables had very short names or were outdated, for example one was `inputMap` from the time where the input was a map instead of a vector.


tmadlener (approved):
This looks good to me. The one thing we could still touch here since we are at it at the moment in any case is to make the generation of the exception messages homogeneous.

Currently it's a mix of 
- Using `fmt::format` in the constructor of the `GaudiException`
- String concatenation using `std::string::operator+` in the constructor of `GaudiException`
- Creating a `std::string` externally (building it up with concatenation) and then passing that to the constructor of the `GaudiException`

But we can also leave that for another time (or when we have some *written* guidelines on how we want to do these things)

PR #341, opened 2025-09-12
Add K4_GAUDI_CHECK macro.

scott-snyder:
Add a macro to reduce repetitive code involved in checking for error return statuses in Gaudi components.

This allows for example to replace code like

```
  if (!m_cellPositionsTool.retrieve()) {
    error() << "Unable to retrieve cell positions tool." << endmsg;
    return StatusCode::FAILURE;
  }
```

with

```
  K4_GAUDI_CHECK( m_cellPositionsTool.retrieve() );
```

BEGINRELEASENOTES
 - Add macro K4_GAUDI_CHECK (in k4FWCore/k4_check.h) to reduce the boilerplate involved in checking Gaudi return status.  If its argument is false, the macro will emit and error and return StatusCode::FAILURE.
ENDRELEASENOTES


k4FWCore/include/k4FWCore/k4_check.h
```diff
@@ -0,0 +1,48 @@
...
+#include "GaudiKernel/StatusCode.h"
+
+/**
+ * @brief Helper to check return codes in a Gaudi component.
+ *
+ * <code>K4_GAUDI_CHECK( EXP )</code> will evaluate @c EXP.  If this results
+ * in something that evaluates to @c false (including @c StatusCode::FAILURE),
+ * we emit an error message and immediately return a @c StatusCode::FAILURE.
+ * This can be used to make checking the results of calls from Gaudi components
+ * easier to read.
```
tmadlener, line 38:
```suggestion
 * <code>K4_GAUDI_CHECK( EXP )</code> will evaluate @c EXP.  If this results
 * in something that evaluates to @c false (including @c StatusCode::FAILURE),
 * we emit an error message and immediately return a @c StatusCode::FAILURE.
 * This can be used to make checking the results of calls from Gaudi components
 * easier to read.
```

Now that you have pointed me to the ATLAS solution and it's comprehensive documentation, I think we should also put the minimal code example this is supposed to replace here.

tmadlener (commented):
> I could add an option to give a custom error message. However,
at ATLAS i've never really felt the need for that (giving the failed expression
along with the line number is generally sufficient), so i'd prefer to keep
this simple to start with and extend later if desired.

Yeah, that makes sense. Usually one of the first thing you do in any case is to grep for the error message and hope for only a handful of results. Also if it allows us to keep boost preprocessor out for now, I am happy.

k4FWCore/include/k4FWCore/k4_check.h
scott-snyder: Thanks.  Extended the documentation along the lines of what we have in ATLAS.

scott-snyder:
Squashed.

jmcarcell:
If this is used in ATLAS and now here, doesn't it make sense that we should have something like this in Gaudi? In some cases we do not always return. I see that in ATLAS every possible one is defined. How does it look in ATLAS, is it only these macros?

scott-snyder:
> If this is used in ATLAS and now here, doesn't it make sense that we should have something like this in Gaudi?

We could consider that, i guess, but using a project-specific name allows for customization later, eg., for error reporting (a la ATLAS).

> In some cases we do not always return. I see that in ATLAS every possible one is defined. How does it look in ATLAS, is it only these macros?

Not quite sure what you're getting at here.  Yes, one sometimes need to be aware that the `*CHECK` macros can return, not so much in `initialize()` (where they're mostly used) but more so in event processing.

`ATH_CHECK `is used throughout Atlas code:

https://acode-browser1.usatlas.bnl.gov/lxr/search?%21v=head&_filestring=&_string=ATH_CHECK

as I said mostly in `initialize()` but in plenty of other places as well.  Other variants such as `*CHECK_RECOVERABLE` and `*CHECK_WITH_CONTEXT` have a few uses, but are dwarfed by the plain `ATH_CHECK`.

jmcarcell:
I had a look at how this is done in LHCb and opened a PR (https://github.com/key4hep/k4FWCore/pull/356) with changes the alternative changes. This alternative is using `StatusCode::andThen` and `StatusCode::orElse`. For your example, it would become:


``` cpp 
return m_cellPositionsTool.retrieve().orElse([&] {
    error() << "Unable to retrieve cell positions tool." << endmsg;
});
```

which is not that verbose and allows chaining multiple operations. It's true that these macros give you more information like the file and line number. I'd rather use built-in functions from Gaudi than macros, however. We could study how to have something similar in Gaudi to print the file and line number for an error.

jmcarcell:
> The existing code would return from initialize() if the call failed,
> while the change prints a message but then ignores the error and continues.

This is a mistake on my side, there should be a `return` at the beginning.

I think the comments are fair, I also noticed it wasn't particularly simple while doing https://github.com/key4hep/k4FWCore/pull/356. So let's go ahead with this. To keep the style that we have in the includes I'll rename it to `GaudiChecks.h` (assuming it is possible that in the future there are more variants of the check, like in ATLAS) unless someone has a preference for some name.

PR #356, opened 2025-11-06
Use .andThen() and orElse() for chaining checks with StatusCodes

jmcarcell:
BEGINRELEASENOTES
- Use .andThen() and orElse() for chaining checks with StatusCodes

ENDRELEASENOTES

related to https://github.com/key4hep/k4FWCore/pull/341.

### DIFF
--- a/k4FWCore/components/IOSvc.cpp
+++ b/k4FWCore/components/IOSvc.cpp
@@ -34,10 +34,12 @@
 #include <tuple>
 
 StatusCode IOSvc::initialize() {
-  Service::initialize().orElse([&]() {
-    error() << "Unable to initialize base class Service." << endmsg;
-    return StatusCode::FAILURE;
-  }).ignore();
+  Service::initialize()
+      .orElse([&]() {
+        error() << "Unable to initialize base class Service." << endmsg;
+        return StatusCode::FAILURE;
+      })
+      .ignore();
 
   if (!m_importedFromk4FWCore) {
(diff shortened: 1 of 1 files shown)


jmcarcell:
I think the first one has the downsides of both. The second is only a minor modification from the standard `orElse`, and I think there would be support to have in Gaudi some message to print the file and line. I'll have a look at that.

> Generally, I like this approach, but it's not entirely trivial to understand for complete newcomers potentially. Especially, if there is a compilation error, this has the potential for quite a bit of confusion, I think.

Yeah, maybe not. In any case it's one way of doing it, one can always do the standard if - else. Some examples can be found in https://gitlab.cern.ch/search?search=orElse&nav_source=navbar&project_id=399&group_id=120&search_code=true&repository_ref=master

test/k4FWCoreTest/src/components/TestUniqueIDGenSvc.cpp
```diff
@@ -18,21 +18,14 @@
...
 
 StatusCode TestUniqueIDGenSvc::initialize() {
-  if (Gaudi::Algorithm::initialize().isFailure()) {
-    return StatusCode::FAILURE;
-  }
-
-  m_service = serviceLocator()->service("UniqueIDGenSvc");
-
-  return StatusCode::SUCCESS;
+  return Gaudi::Algorithm::initialize().andThen([&]() { m_service = serviceLocator()->service("UniqueIDGenSvc"); });
```
m-fila: We should check that  `m_service` isn't `nullptr`. The check is missing in the original version too

Issue #225, opened 2024-08-20
`EventContext` in k4FWCore and functional algorithms


m-fila:
In Gaudi functional the algorithms' `operator()` can accept an `EventContext` that will be provided by the scheduler. In k4FWCore functional algorithms it's not allowed. Both Gaudi and k4FWCore non-functional algorithms accept `EventContext` in their `execute`

The `EventContext` holds information about:
- a concurrent slot used by the event,
- event number (ordinal number given by scheduler)
- id consisting of event id and run id (to be set from event data).
In the k4FWCore the standard way of supplying event id and run id is with an `EventHeaderCollection` data object so this is not used currently

The `EventContext` information could be still useful for less-standard algorithms like `EventHeaderCreator` #224. The `EventContext` run id and event id is used in some logging statements but is never set in k4FWCore so a dummy (uninitialized?) value is used

This is not a big problem and rather invisible during "standard" usage. I create this item mostly to keep track of differences between standard Gaudi facilities and k4FWCore

PR #360, opened 2025-11-17
Add support for passing EventContext in functional algorithms

jmcarcell:
BEGINRELEASENOTES
- Add support for passing EventContext in functional algorithms
- Add a test consumer, transformer and multitransformer that take an `EventContext` and tests that run them

ENDRELEASENOTES
  
This wasn't added initially because then the implementation was a bit simpler than tackling all at once. Fix https://github.com/key4hep/k4FWCore/issues/225.

### DIFF
--- a/k4FWCore/include/k4FWCore/FunctionalUtils.h
+++ b/k4FWCore/include/k4FWCore/FunctionalUtils.h
@@ -171,12 +171,17 @@ namespace details {
   // pack expansion is needed. Once to filter out the event context, and the other one to
   // create the tuple of vectors of handles
   template <template <typename> class Handle, typename Tuple>
-  struct to_tuple_of_handle_vec;
+  struct tuple_of_handle_vec {
+    using type = std::tuple<>;
+  };
 
   template <template <typename> class Handle, typename... Ts>
-  struct to_tuple_of_handle_vec<Handle, std::tuple<Ts...>> {
+  struct tuple_of_handle_vec<Handle, std::tuple<Ts...>> {
     using type = std::tuple<std::vector<Handle<typename EventStoreType<Ts>::type>>...>;
   };
+
+  template <template <typename> class Handle, typename... Ts>
+  using tuple_of_handle_vec_t = typename tuple_of_handle_vec<Handle, Ts...>::type;
 
   template <size_t Index, typename... In, typename... Handles, typename InputTuple>
   void readVectorInputs(const std::tuple<Handles...>& handles, auto thisClass, InputTuple& inputTuple) {
--- a/k4FWCore/include/k4FWCore/Consumer.h
+++ b/k4FWCore/include/k4FWCore/Consumer.h
@@ -60,7 +60,7 @@ namespace details {
     template <typename T>
     using InputHandle_t = Gaudi::Functional::details::InputHandle_t<Traits_, T>;
 
-    to_tuple_of_handle_vec<InputHandle_t, filter_evtcontext_t<In...>>::type m_inputs;
+    tuple_of_handle_vec_t<InputHandle_t, filter_evtcontext_t<In...>> m_inputs;
 
     std::array<Gaudi::Property<DataObjID>, N_in> m_inputLocationsSingle;
     std::array<Gaudi::Property<std::vector<DataObjID>>, N_in> m_inputLocationsVector;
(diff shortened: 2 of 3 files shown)


tmadlener (commented):
For my understanding: Are some of the changes here because upstream Gaudi changed them similarly? Or is all of this necessary to get the `EventContext` to the Functional algorithms?

k4FWCore/include/k4FWCore/FunctionalUtils.h
```diff
@@ -140,6 +142,42 @@ namespace details {
     }
   };
 
+  template <typename... In>
+  struct filter_evtcontext<EventContext, In...> {
+    static_assert(!std::disjunction_v<std::is_same<EventContext, In>...>,
+                  "EventContext can only appear as first argument");
+
+    using type = std::tuple<In...>;
+    static constexpr std::size_t size = std::tuple_size<type>::value;
```
tmadlener:
```suggestion
    static constexpr std::size_t size = std::tuple_size_v<type>;
```

test/k4FWCoreTest/src/components/ExampleFunctionalConsumerEventContext.cpp
```diff
@@ -0,0 +1,95 @@
...
+        throw std::runtime_error(error.str());
+      }
+      i++;
+    }
+    {
+      std::lock_guard<std::mutex> lock(m_mutex);
+      if (m_eventNumbersSeen.find(ctx.evt()) != m_eventNumbersSeen.end()) {
+        throw std::runtime_error("Event number " + std::to_string(ctx.evt()) + " seen multiple times");
+      }
+      m_eventNumbersSeen.insert(ctx.evt());
```
tmadlener:
```suggestion
        m_eventNumbersSeen.insert(ctx.evt());
      }
```
The insertion should also be locked, I think.

test/k4FWCoreTest/src/components/ExampleFunctionalTransformerEventContext.cpp
```diff
@@ -0,0 +1,88 @@
...
+      new_particle.setTime(particle.getTime() + m_offset);
+      new_particle.setMass(particle.getMass() + m_offset);
+    }
+
+    {
+      std::lock_guard<std::mutex> lock(m_mutex);
+      if (m_eventNumbersSeen.find(ctx.evt()) != m_eventNumbersSeen.end()) {
+        throw std::runtime_error("Event number " + std::to_string(ctx.evt()) + " seen multiple times");
+      }
+      m_eventNumbersSeen.insert(ctx.evt());
```
tmadlener:
```suggestion
        m_eventNumbersSeen.insert(ctx.evt());
      }
```

test/k4FWCoreTest/options/ExampleFunctionalMemory.py
tmadlener: More of a general comment / question to k4FWCore, mainly because I saw the explicit guarding of mutable state below: Do we run some tests (or these tests) on multiple threads?

tmadlener (commented):
OK. After having had a closer look, it seems like some of the diff size is also explained by the fact, that things are just moving around quite a bit and it's not actually that many changes (that initially) looked unrelated.

I have left a few minor comments below.

k4FWCore/include/k4FWCore/FunctionalUtils.h
```diff
@@ -140,6 +142,42 @@ namespace details {
...
+  // This is a helper class to create the input and output types because a double parameter
+  // pack expansion is needed. Once to filter out the event context, and the other one to
+  // create the tuple of vectors of handles
+  template <template <typename> class Handle, typename Tuple>
+  struct wrap_tuple_types;
+
+  template <template <typename> class Handle, typename... Ts>
+  struct wrap_tuple_types<Handle, std::tuple<Ts...>> {
+    using type = std::tuple<std::vector<Handle<typename EventStoreType<Ts>::type>>...>;
+  };
```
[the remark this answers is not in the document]
tmadlener: What about `transform_to_handle_vec` as an alternative name? That is what it does in the end, as the result of `filter_evtcontext_t<In...>` is what is being transformed here, right?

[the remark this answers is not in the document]
tmadlener: I like that. But I would keep the `::type` to keep it similar(ish) to what is usually done with these traits.

edit: Or alternatively (equally long) to keep it similar to `filter_evtcontext_t`: `tuple_of_handle_vec_t`(?)

PR #361, opened 2025-11-19
Only set the OutputLevel from the command line if one was passed

tmadlener:
BEGINRELEASENOTES
- Make sure to only set an `OutputLevel` on the `ApplicationMgr` if `--log-level` has actually been used instead of setting it to `INFO` (regardless of the user configuration in the option file) even if no `--log-level` has passed

ENDRELEASENOTES

This was introduced in #306, from where also the original bug stems. Without this fix even if a user has `OutputLevel=DEBUG` in their option file, we have later reset it to `INFO` because that was the default value for `--log-level`. We should only override values from the option file if users explicitly pass another `--log-level` via the CLI.

### DIFF
--- a/k4FWCore/scripts/k4run
+++ b/k4FWCore/scripts/k4run
@@ -8,6 +8,8 @@ import signal
 import warnings
 from pathlib import Path
 
+from Gaudi import Configuration
+
 from k4FWCore.utils import load_file, get_logger, set_log_level, LOG_LEVELS
 
 # Terminate the process if a SIGPIPE is received
@@ -213,8 +215,6 @@ def main():
     set_log_level(opts[0].log_level or "INFO")
 
     if opts[0].list:
-        from Gaudi import Configuration
-
         cfgdb = Configuration.cfgDb
         logger.info("Available components:\n")
         for item in sorted(cfgdb):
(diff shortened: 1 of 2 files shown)


k4FWCore/scripts/k4run
```diff
@@ -267,10 +267,16 @@ def main():
         else:
             conf.setProp(optionName.rsplit(".", 1)[1], opts_dict[optionName])
 
-    if opts.verbose:
-        from Gaudi.Configuration import VERBOSE
+    if opts.log_level:
+        try:
```
jmcarcell: Alternatively, which levels are available can be found in 
``` python
>>> from GaudiKernel import Constants
>>> print(Constants.__all__)
['ALL',
 'VERBOSE',
 'DEBUG',
 'INFO',
 'WARNING',
 'ERROR',
 'FATAL',
 'TRUE',
 'FALSE']
```
(there is a comment about removing `TRUE` and `FALSE` since, at least, 2009...). I'm not sure if this is better than the try - except.

jmcarcell: > I could use the `Constants.__all__` here, but than I would still have to clean them up a bit first, I think. Or alternatively, introduce more names for the python log levels to match the ones that come from Gaudi.

As long as `LOG_LEVELS` is a subset of the constants nothing needs to be done. What I meant is simply changing the try - except to looking in `Constants.__all__` and failing if not found.

PR #351, opened 2025-10-22
Make KeepDropSwitch fail earlier on invalid inputs

tmadlener:
BEGINRELEASENOTES
- Check the commands for the `KeepDropSwitch` in the constructor instead of re-parsing them every time `isOn` is called. 
- Move the `KeepDropSwitch` into the `k4FWCore` namespace and deprecate the unnamespaced usage.
- Make the `IOSvc` fail in `initialize` in case the `outputCommands` are invalid
- Add unittests for `KeepDropSwitch`

ENDRELEASENOTES

Not sure if the deprecation of the global namespace definition of `KeepDropSwitch` is necessary or if we can simply remove the deprecated definition immediately.

Fixes #348

PR #362, opened 2025-11-20
Make it possible to keep or drop all collections of a given datatype

tmadlener:
BEGINRELEASENOTES
- Introduce the `type` subcommand to the `outputCommands` grammar to keep / drop all collections of a given datatype.

ENDRELEASENOTES

This functionality was also present for LCIO / Marlin and I think it's quite useful, because it just allows one to write e.g. `drop type edm4hep::SimTrackerHitCollection` to drop all sim tracker hits without having to know all names first.

I have implemented this here instead of podio, because I think it fits better here.

For now I have also kept the overload of `KeepDropSwitch::isOn` that only takes a single string, but I don't think there are any users outside for which backwards compatibility would be a concern (see also #351 where the `KeepDropSwitch` was moved into the `k4FWCore` namespace). Removing that would require touching the existing unit tests, but that would not be a real problem, I think.

There is also some initial refactoring of the `KeepDropSwitch` that could in principle be pulled into a separate PR if necessary.

### DIFF
--- a/doc/PodioInputOutput.md
+++ b/doc/PodioInputOutput.md
@@ -183,6 +183,8 @@ io_svc.outputCommands = [
 ]
 ```
 
+The type name here is matched against the type obtained with `podio::CollectionBase::getTypeName()`, i.e. *Collection* has to be included. Only exact matches are considered for the application of the command.
+
 :::{note}
 The commands are processed in order so the last relevant *keep* or *drop* for any given collection will decide.
 :::


tmadlener:
I am fairly happy with the syntax / grammar for the output commands here. The main question I still have is whether it should be `keep type XYZCollection` or whether we shorten it to only `keep type XYZ`. The main problem with that is that `podio::UserDataCollection` become only, e.g. `float`, or `int` which looks slightly weird, IMO.

jmcarcell:
> I am fairly happy with the syntax / grammar for the output commands here. The main question I still have is whether it should be `keep type XYZCollection` or whether we shorten it to only `keep type XYZ`. The main problem with that is that `podio::UserDataCollection` become only, e.g. `float`, or `int` which looks slightly weird, IMO.

I would say, leave it as it is now, with the namespace.

doc/PodioInputOutput.md
```diff
@@ -175,6 +175,19 @@ k4run <steering-file> --IOSvc.outputCommands \
...
 ::::
 
+It is also possible to specify entire datatypes for keeping or dropping via the `type` sub-command, e.g. to keep everything but any collection that is of type `edm4hep::SimTrackerHitCollection` one would do:
+
+```python
+io_svc.outputCommands = [
+    "drop type edm4hep::SimTrackerHitCollection",
+]
+```
+
```
jmcarcell:
```suggestion
The type obtained is matched against the type obtained with `coll.getTypeName()` and only if an exact match is found the rule is used.

```
Maybe something like this

jmcarcell (approved):

PR #363, opened 2025-11-20
Use fixtures consistently for test dependencies

tmadlener:
BEGINRELEASENOTES
- Use cmake fixtures to properly denote dependencies between tests

ENDRELEASENOTES

This is mostly introducing new fixtures where we previously had `DEPENDS`. I did some minor "cleanup" by lifting a few dedicated `set_test_properties` into the `add_test_fwcore` call.

I have run 
```bash
for i in {0..9}; do 
  rm -rf build/test/k4FWCoreTest/*.root
  ctest --test-dir build -j10 --schedule-random 2>&1 >/dev/null && echo "$i: SUCCESS"; 
done 
```
and that worked without issues for me, so I suppose dependencies should be set correctly.

### DIFF
--- a/test/k4FWCoreTest/CMakeLists.txt
+++ b/test/k4FWCoreTest/CMakeLists.txt
@@ -110,7 +110,12 @@ function(add_test_fwcore testname)
     set_tests_properties(${testname} PROPERTIES ${TEST_PROPERTIES})
   endif()
   if(ADD_TO_CHECK_FILES)
-    set_tests_properties(${testname} PROPERTIES FIXTURES_SETUP TestFiles)
+    get_test_property(${testname} FIXTURES_SETUP existing_fixtures)
+    if(existing_fixtures AND NOT existing_fixtures STREQUAL "NOTFOUND")
+      set_tests_properties(${testname} PROPERTIES FIXTURES_SETUP "${existing_fixtures};TestFiles")
+    else()
+      set_tests_properties(${testname} PROPERTIES FIXTURES_SETUP TestFiles)
+    endif()
   endif()
 endfunction()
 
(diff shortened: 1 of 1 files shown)


test/k4FWCoreTest/CMakeLists.txt
```diff
@@ -237,17 +241,17 @@ set_tests_properties(CheckLoadedFileCorrectPathOnError
   PROPERTIES PASS_REGULAR_EXPRESSION [=[  File ".*/test/k4FWCoreTest/options/checkLoadedFileProperties.py", line 36, in <module>]=]
 )
 
-add_test_fwcore(ParticleIDMetadataFramework options/ExampleParticleIDMetadata.py)
+add_test_fwcore(ParticleIDMetadataFramework options/ExampleParticleIDMetadata.py PROPERTIES FIXTURES_SETUP ParticleIDMetadataFile)
 
-add_test_fwcore(EfficiencyFilter options/ExampleEfficiencyFilter.py PROPERTIES DEPENDS EventHeaderFiller FIXTURES_REQUIRED ProducerFile ADD_TO_CHECK_FILES)
+add_test_fwcore(EfficiencyFilter options/ExampleEfficiencyFilter.py PROPERTIES FIXTURES_REQUIRED "ProducerFile\\\\;EventHeaderFillerFile" ADD_TO_CHECK_FILES)
```
jmcarcell: Because of escaping, this also works with three `\\\`...

jmcarcell:
> This is mostly introducing new fixtures where we previously had `DEPENDS`. I did some minor "cleanup" by lifting a few dedicated `set_test_properties` into the `add_test_fwcore` call.
I think I never bothered for some tests since if they don't fail then `DEPENDS` and fixtures work the same way :smiley: 

One day randomization can be added to the key4hep builds but now I'm not very interested as it allows me to compare before and after some changes.

tmadlener:
Yeah without randomization `DEPENDS` and fixtures do pretty much the same, but in many of our tests we have a test that writes something and than another that reads that file. With `DEPENDS` re-running the reading part doesn't automatically trigger the writing part but with fixtures I will always get a fresh output from the write part. That is quite handy if you need to change both tests as it then makes sure you are not failing the reading part because you are still working with an old output from the writing part.

> One day randomization can be added to the key4hep builds

I almost opened a PR to key4hep-actions. I think we would maybe have to do a similar thing as with the cmake extra args, to make it possible for repositories to opt-in as I don't think randomizing test order will work in many of our repositories at the moment.

PR #371, opened 2026-01-16
Make sure we can consume EDM4hep 1.0

tmadlener:
BEGINRELEASENOTES
- Ensure we can still configure and build with EDM4hep 1.0

ENDRELEASENOTES


tmadlener:
Downstream build seems to fail for unrelated reasons.

PR #372, opened 2026-01-16
Add Gaudi Functional C++ Class Generator

ianna:
BEGINRELEASENOTES

 - Gaudi Functional C++ Class Generator (`generateFunctional`)

ENDRELEASENOTES

# Gaudi Functional C++ Class Generator

`k4FWCore/helpers/generateFunctional` generates Gaudi Functional C++ algorithm boilerplate based on user-defined specifications. It supports both **Gaudi::Functional** and **k4FWCore** frameworks.

## Features

- **Framework support**: Both Gaudi and k4FWCore (default: k4FWCore)
- **5 functional types**: `Consumer`, `Producer`, `Transformer`, `MultiTransformer`, `FilterPredicate`
- **Type inference**: functional type is auto-detected from the number of inputs and outputs; only `filter` needs to be specified explicitly
- **Smart type parsing**: handles complex types like `podio::UserDataCollection<float>`
- **Constructor scaffolding**: generates `KeyValue`/`KeyValues` initialization with correct brace wrapping for all functional types
- **Type-safe `operator()`**: creates the correct signature and a stub implementation
- **Gaudi properties**: add configurable `Gaudi::Property<T>` members with optional `private:` placement (`--private-properties`)
- **EDM4hep/podio support**: automatic `#include` directives for edm4hep and podio types
- **Runtime (variable-length) collections**: `--runtime-inputs` and `--runtime-outputs` for `std::vector<const T*>` / `std::vector<T>` patterns
- **Type aliases**: `--type-aliases` emits `using XxxColl = …;` aliases and uses them in the signature
- **Event context**: `--event-context` adds `const EventContext&` as the first `operator()` argument and scaffolds `finalize()` with mutex bookkeeping
- **CMake skeleton**: `--cmake` also writes a `CMakeLists.txt` next to the source
- **Safe writes**: refuses to overwrite existing files unless `--force` is passed
- **Command tracking**: records the generation command as a comment in the output file

## Input format

For inputs/outputs: `TypeName:LocationKey`
- **TypeName**: required (supports templates like `podio::UserDataCollection<float>`)
- **LocationKey**: optional — defaults to the type name stripped of namespace and `Collection` suffix

For properties: `Type:Name:DefaultValue[:Description]`

### k4FWCore Producer (type inferred)
```bash
generateFunctional MyProducer \
  -o 'edm4hep::MCParticleCollection:OutputCollection' \
  -p 'int:ExampleInt:3:Example integer property'
```

### k4FWCore Multi-output Producer (MultiTransformer inferred from multiple outputs)
```bash
generateFunctional MyProducer \
  -o 'podio::UserDataCollection<float>:VectorFloat' \
     'edm4hep::MCParticleCollection:MCParticles1' \
     'edm4hep::SimTrackerHitCollection:SimTrackerHits' \
  -p 'int:ExampleInt:3:Example int that can be used in the algorithm' \
     'int:magicNumberOffset:0:Integer to add to the dummy values'
```

### k4FWCore Consumer (type inferred)
```bash
generateFunctional MyConsumer \
  -i 'edm4hep::MCParticleCollection:InputCollection'
```

### k4FWCore Transformer (type inferred, properties private)
```bash
generateFunctional MyTransformer \
  -i 'edm4hep::MCParticleCollection:InputCollection' \
  -o 'edm4hep::MCParticleCollection:OutputCollection' \
  --private-properties \
  -p 'int:Offset:10:Integer to add to values'
```

### Runtime (variable-length) collections
```bash
generateFunctional MyConsumer \
  -i 'edm4hep::MCParticleCollection:Inputs' \
  --runtime-inputs 'edm4hep::MCParticleCollection:Inputs:MCParticles0,MCParticles1'
```

### With EventContext and finalize() scaffold
```bash
generateFunctional MyTransformer \
  -i 'edm4hep::MCParticleCollection:InputCollection' \
  -o 'edm4hep::MCParticleCollection:OutputCollection' \
  --event-context --private-properties
```

### FilterPredicate (type must be explicit)
```bash
generateFunctional MyFilter filter \
  -i 'edm4hep::MCParticleCollection:InputCollection'
```

### Gaudi framework, namespace, CMake output
```bash
generateFunctional MySum \
  -i 'Input1:Loc1' 'Input2:Loc2' \
  -o 'Output:OutLoc' \
  --framework gaudi -n MyNamespace --cmake
```

## Command-line options

```
positional arguments:
  class_name                       Name of the C++ class to generate
  {consumer,producer,transformer,filter}
                                   Functional type (optional — inferred when omitted)
options:
  -i, --inputs TYPE:KEY [...]      Input collection specs
  -o, --outputs TYPE:KEY [...]     Output collection specs (mutually exclusive with --runtime-outputs)
  --runtime-outputs TYPE           Dynamic output: returns std::vector<TYPE> (k4FWCore only)
  --runtime-inputs TYPE:KEY:DEF[,DEF2,...] [...]
                                   Variable-length inputs with default location names (k4FWCore only)
  -p, --properties TYPE:NAME:DEFAULT[:DESC] [...]
                                   Gaudi::Property<T> members
  -n, --namespace NAME             Wrap class in a C++ namespace
  --framework {gaudi,k4fwcore}     Target framework (default: k4fwcore)
  --use-class                      Emit 'class … { public: }' instead of 'struct'
  --private-properties             Place properties under 'private:'
  --type-aliases                   Emit 'using XxxColl = …;' aliases for input types
  --all-keyvalues                  Treat all inputs as KeyValues vectors
  --keyvalues-inputs KEY [...]     Promote specific inputs to KeyValues vectors
  --event-context                  Add EventContext arg and finalize() scaffold
  --cmake                          Also emit a CMakeLists.txt skeleton
  -f, --output-file PATH           Output file (default: <ClassName>.cpp)
  --force                          Overwrite existing files
```

## How to run

```bash
# Recommended — uv resolves Python and Jinja2 automatically (PEP 723)
uv run generateFunctional MyProducer -o 'edm4hep::MCParticleCollection:MCParticles'

# Direct execution via shebang (requires uv on PATH)
./generateFunctional MyProducer -o 'edm4hep::MCParticleCollection:MCParticles'

# Plain Python (requires jinja2 in the active environment)
python3 generateFunctional MyProducer -o 'edm4hep::MCParticleCollection:MCParticles'
```

After `cmake --install`, `generateFunctional` is available on `PATH` in the Key4hep environment.


jmcarcell:
Without any tests, inevitably this will not work in the future if something changes and we won't be able to know.

k4FWCore/helpers/gaudi_gen.py
```diff
@@ -0,0 +1,489 @@
...
+    
+    # k4FWCore uses struct by default, Gaudi uses class by default
+    if framework == 'k4fwcore':
+        class_keyword = "class" if use_class else "struct"
+        public_keyword = "public:\n" if use_class else ""
+    else:
+        class_keyword = "struct" if not use_class else "class"
+        public_keyword = "" if not use_class else "public:\n"
+    
+    code = f"""// Generated by Gaudi Functional C++ Class Generator
```
andresailer: use textwrap.dedent
https://docs.python.org/3/library/textwrap.html

k4FWCore/helpers/gaudi_gen.py
```diff
@@ -0,0 +1,489 @@
...
+                       help='Type of functional to generate')
+    parser.add_argument('-i', '--inputs', nargs='*', default=[],
+                       help='Input data specifications (format: "Type:Location" or just "Type")')
+    parser.add_argument('-o', '--outputs', nargs='*', default=[],
+                       help='Output data specifications (format: "Type:Location" or just "Type")')
+    parser.add_argument('-n', '--namespace', default='',
+                       help='Namespace for the class')
+    parser.add_argument('-f', '--output-file', 
+                       help='Output file name (default: <ClassName>.cpp)')
+    parser.add_argument('--framework', choices=['gaudi', 'k4fwcore'], default='k4fwcore',
```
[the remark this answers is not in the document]
ianna: > Why do we need the gaudi framework option in k4fwcore?

@andresailer - thanks for reviewing the PR! I thought since the README mentions the Gaudi tutorial it might be useful :-)

tmadlener (commented):
Thanks a lot for this. I think the general idea of the script goes into the exact right direction, i.e. we want something that generates boilerplate for something that the user wants. I didn't look in extreme detail, but I see a few general improvements (also partially highlighted in the inline comments):
- Currently the user needs to know which functional type they want. But this is entirely specified by the number of inputs and outputs, so I would just determine that from there.
- Generally the implementation could probably be improved by sprinkling a few classes into the whole thing for holding intermediate information instead of passing around tuples of strings of various lengths. This would probably also make it easier to simply do some of the parsing / processing up-front and then pass the information around instead of re-parsing it several times.
- The current implementation does not handle the possibility of variable length inputs / outputs (I think). This is only possible in `k4FWCore` Functional algorithms though.

I think this is borderline complex enough to warrant the use of a template engine to handle all the string formatting. It introduces some overhead, and would require writing some templates, but it would simplify the python script parts by quite potentially.

As already mentioned we definitely need tests that ensure that the outputs compile. That would probably be the first thing I do, because once that is in place refactoring and extending the implementation can be done with some guard rails.

k4FWCore/helpers/gaudi_gen.py
```diff
@@ -0,0 +1,837 @@
...
+k4FWCore and native Gaudi::Functional frameworks.
+
+Refactoring notes (addressing PR #372 review by tmadlener):
+  - All CLI strings are parsed up-front into typed dataclasses (DataSpec,
+    RuntimeInputSpec, PropertySpec, AlgorithmSpec).  No re-parsing inside
+    generator methods.
+  - The functional type is inferred from the number of inputs / outputs when
+    not supplied explicitly on the command line.
+  - All C++ output is produced through Jinja2 templates, keeping Python logic
+    and string layout cleanly separated.
```
tmadlener:
```suggestion
```
I don't think we need the history of how we came to the final thing in the end.

k4FWCore/helpers/gaudi_gen.py
```diff
@@ -0,0 +1,837 @@
...
+"""
+
+from __future__ import annotations
```
tmadlener:
```suggestion
```
If this is necessary we should probably find a workaround, because this is superseeded. [PEP 563](https://peps.python.org/pep-0563/). See also: https://www.reddit.com/r/learnpython/comments/1psteqc/comment/nvcfeoh/

Doesn't have to be in this PR though in case it's too much work. Python 3.14 should solve most issues related to this trivially: [deferred annotations](https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-deferred-annotations)

k4FWCore/helpers/gaudi_gen.py
```diff
@@ -0,0 +1,837 @@
...
+      --runtime-inputs 'edm4hep::MCParticleCollection:Inputs:MCParticles0,MCParticles1'
+""",
+    )
+    parser.add_argument("class_name", help="Name of the C++ class to generate")
+    parser.add_argument(
+        "functional_type", nargs="?",
+        choices=["consumer", "producer", "transformer", "filter"],
+        help="Functional type (inferred from I/O counts when omitted)",
+    )
+    parser.add_argument("-i", "--inputs",  nargs="*", default=[], metavar="TYPE:KEY")
```
tmadlener: Can all of the arguments get a `help` here? For some of them it's kind of obvious, but for others I currently would have trouble to tell what it should be.

k4FWCore/helpers/gaudi_gen.py
```diff
@@ -0,0 +1,837 @@
...
+    parser.add_argument("--cmake", action="store_true", default=False)
+    return parser
+
+
+def main() -> None:
+    parser = _build_parser()
+    args = parser.parse_args()
+
+    if args.runtime_outputs and args.outputs:
+        parser.error("--runtime-outputs and --outputs are mutually exclusive.")
```
tmadlener: This could be done via a `mutually_exclusive_group`, I think.

k4FWCore/helpers/gaudi_gen.py
```diff
@@ -0,0 +1,837 @@
...
+    if args.runtime_outputs and args.framework != "k4fwcore":
+        parser.error("--runtime-outputs is only supported with --framework k4fwcore.")
+
+    try:
+        spec = _build_spec(args)
+    except ValueError as exc:
+        parser.error(str(exc))
+        return
```
tmadlener:
```suggestion
        sys.exit(1)
```
Makes it easier to catch errors in scripts and CI for example

k4FWCore/helpers/gaudi_gen.py
```diff
@@ -0,0 +1,837 @@
...
+  # k4FWCore producer (type inferred)
+  gaudi_gen.py MyProducer -o 'edm4hep::MCParticleCollection:MCParticles'
+
+  # k4FWCore multi-output producer with properties
+  gaudi_gen.py MyProducer \\
+      -o 'edm4hep::MCParticleCollection:MCParticles' \\
+         'edm4hep::TrackCollection:Tracks' \\
+      -p 'int:ExampleInt:3:An example integer property'
+
+  # Gaudi transformer with namespace (type inferred)
```
tmadlener: What does *namespace* here refer to? (I guess this goes in hand with `help` messages for the arguments).

tmadlener (commented):
As mentioned in the meeting today already it would be nice to have tests for this. Given that you also generate a minimal cmake script this should be fairly straight forward, essentially for various options (I would say each type of algorithm at least once) one would 
- Run the script (probably inside a "sandbox" directory) to generate source code and cmake
- Run cmake
- Run build

The main problem with that is that we would have to hook that into the existing test harness in the `CMakeLists.txt` if we just want to run it as part of the "standard" CI. An alternative option would be a custom github actions workflow where we could run this without that harness. (Opinions @jmcarcell?).

In any case I think a first step would to just write some bash scripts for each option with the three steps above and then we can decide later where / how we hook them up to CI).

---------

Some other general comments:
- `gaudi_gen.py` is probably not descriptive enough. In the end people would like to use this from within a Key4hep environment and I think something like `generatFunctional.py` or `generateAlgorithm.py` (or even without the `.py` suffix) would make that a bit more explicit
- This needs to be installed for it to be truly useful, otherwise people will have to clone k4FWCore and get the script from there.

For that something similar to what we do for `k4run`:
https://github.com/key4hep/k4FWCore/blob/dd8d8919348c2b576aa4062deaa5c2bf9bb8ff14/k4FWCore/CMakeLists.txt#L20

In this case it would probably have to be something like
```cmake
gaudi_install(SCRIPTS helpers)
```

or alternatively via an explicit call to cmake `install`.

jmcarcell:
I tried this and there are several options that are not explained, then I tried them and they seem not to do anything. Like `--all-keyvalues`, or `--type-aliases`.  

> The main problem with that is that we would have to hook that into the existing test harness in the CMakeLists.txt if we just want to run it as part of the "standard" CI. An alternative option would be a custom github actions workflow where we could run this without that harness. (Opinions @jmcarcell?).
Hmm if it doesn't run in ctest then it's not tested regularly, but things don't change that much here so that it would stop working. If it's going to use the cmake it creates then I think it has to be outside because it calls `find_package(k4FWCore)` that won't be there until `k4FWCore` is installed, and probably is more similar to how people will use it in the future. In that case we can have a different workflow in `.github/workflows` that is unrelated to the build one, that may also build `k4FWCore`. 

I wonder at which point, in a repo that already has algorithms, it's just easier to copy and paste an existing algorithm than get the command right to run the script. If there are many different times one has to copy and paste all of them with all the `edm4hep::`. Maybe `edm4hep::` could be removed and assume it's an EDM4hep collection (true for now, but probably not in the future). Also one has to remember the colon separators:
```
python test.py MyProducer -o 'edm4hep::MCParticleCollection:MCParticles' 'edm4hep::TrackCollection:Tracks' -i "edm4hep::MCParticleCollection" -p 'int:ExampleInt:3:An example integer property'
```

k4FWCore/helpers/gaudi_gen.py
ianna: Good point - `help` added, thanks!

k4FWCore/helpers/gaudi_gen.py
```diff
@@ -0,0 +1,489 @@
...
+    help_text += "  # Gaudi transformer (generates class by default)\n"
+    help_text += "  python gaudi_gen.py MyTransformer transformer \\\n"
+    help_text += "    -i 'InputType:InputLoc' \\\n"
+    help_text += "    -o 'OutputType:OutputLoc' \\\n"
+    help_text += "    --framework gaudi\n"
+    
+    return help_text
+
+
+def parse_data_spec(spec: str) -> Tuple[str, str]:
```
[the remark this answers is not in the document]
ianna: refactored, thanks!

k4FWCore/helpers/gaudi_gen.py
```diff
@@ -0,0 +1,489 @@
...
+    elif functional_type == 'transformer':
+        in_sig = ', '.join([f"const {t}&" for t in in_types]) if in_types else ""
+        if len(out_types) == 1:
+            return f"{out_types[0]}({in_sig})"
+        else:
+            out_sig = ', '.join(out_types)
+            return f"std::tuple<{out_sig}>({in_sig})"
+    elif functional_type == 'filter':
+        in_sig = ', '.join([f"const {t}&" for t in in_types])
+        return f"bool({in_sig})"
```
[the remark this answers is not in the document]
ianna: done, thanks!

k4FWCore/helpers/gaudi_gen.py
```diff
@@ -0,0 +1,489 @@
...
+            return f"std::tuple<{', '.join(out_types)}> operator()() const override"
+    elif functional_type == 'transformer':
+        params = ', '.join([f"const {t}& in{i+1}" for i, t in enumerate(in_types)])
+        if len(out_types) == 1:
+            return f"{out_types[0]} operator()({params}) const override"
+        else:
+            return f"std::tuple<{', '.join(out_types)}> operator()({params}) const override"
+    elif functional_type == 'filter':
+        params = ', '.join([f"const {t}& in{i+1}" for i, t in enumerate(in_types)])
+        return f"bool operator()({params}) const override"
```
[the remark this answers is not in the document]
ianna: done, thanks!

k4FWCore/helpers/gaudi_gen.py
ianna: done, thanks!

k4FWCore/helpers/gaudi_gen.py
ianna: this is unreachable line - removed and added a comment, thanks

k4FWCore/helpers/gaudi_gen.py
ianna: added comments and help to explain it, thanks

k4FWCore/helpers/gaudi_gen.py
ianna: thanks, done!

ianna:
> I tried this and there are several options that are not explained, then I tried them and they seem not to do anything. Like `--all-keyvalues`, or `--type-aliases`.
> 
> > The main problem with that is that we would have to hook that into the existing test harness in the CMakeLists.txt if we just want to run it as part of the "standard" CI. An alternative option would be a custom github actions workflow where we could run this without that harness. (Opinions @jmcarcell?).
> > Hmm if it doesn't run in ctest then it's not tested regularly, but things don't change that much here so that it would stop working. If it's going to use the cmake it creates then I think it has to be outside because it calls `find_package(k4FWCore)` that won't be there until `k4FWCore` is installed, and probably is more similar to how people will use it in the future. In that case we can have a different workflow in `.github/workflows` that is unrelated to the build one, that may also build `k4FWCore`.
> 
> I wonder at which point, in a repo that already has algorithms, it's just easier to copy and paste an existing algorithm than get the command right to run the script. If there are many different times one has to copy and paste all of them with all the `edm4hep::`. Maybe `edm4hep::` could be removed and assume it's an EDM4hep collection (true for now, but probably not in the future). Also one has to remember the colon separators:
> 
> ```
> python test.py MyProducer -o 'edm4hep::MCParticleCollection:MCParticles' 'edm4hep::TrackCollection:Tracks' -i "edm4hep::MCParticleCollection" -p 'int:ExampleInt:3:An example integer property'
> ```

Hi @jmcarcell Please check added k4FWCore/helpers/README.md and also help on the options. Thanks

ianna:
<html><head></head><body><p>Here's what was created:</p>
<p><strong><code>k4FWCore/helpers/tests/</code></strong> — 8 individual test scripts + shared infrastructure:</p>

Script | Covers
-- | --
test_producer.sh | Single output, property
test_consumer.sh | Single input, property
test_transformer.sh | Single in/out, --private-properties
test_multitransformer.sh | Multiple in/out, --type-aliases, podio::UserDataCollection
test_filter.sh | FilterPredicate
test_runtime_consumer.sh | --runtime-inputs / KeyValues
test_runtime_transformer.sh | --runtime-outputs / std::vector<T> return
test_event_context.sh | --event-context + finalize() scaffold
test_gaudi_framework.sh | --framework gaudi + --namespace
_test_common.sh | Shared: locates gaudiGen.py (installed or source), creates sandbox, run_cmake_build()
run_all_tests.sh | Runs all test_*.sh, reports pass/fail summary


<p><strong><code>k4FWCore/CMakeLists.txt</code></strong> — added:</p>
<pre><code class="language-cmake">install(PROGRAMS helpers/gaudiGen.py
  DESTINATION ${CMAKE_INSTALL_BINDIR})
</code></pre>
<p>This puts <code>gaudiGen.py</code> on <code>PATH</code> in a Key4hep environment, so the test scripts (and users) can call it without knowing where k4FWCore was cloned. The <code>_test_common.sh</code> helper checks the installed location first and falls back to the source tree.</p></body></html>

ianna:
@jmcarcell and @andresailer - please see instructions on how to run the tests [here](https://github.com/ianna/k4FWCore/blob/721b470f44701b359402897cbe87cd32ee6cf98c/k4FWCore/helpers/tests/README.md) -- the README.md is a part of this PR. Please let me know if there is anything else needed. Thanks.

ianna:
@jmcarcell and @andresailer - please review. Thanks!

ianna:
@jmcarcell and @andresailer - please review. Thanks!

k4FWCore/helpers/tests/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+The tests run entirely in bash — no CTest or Python test framework needed.
+Each script generates a C++ algorithm, runs `cmake`, and compiles it.
+
+---
+
+## Prerequisites
+
+A Key4hep environment and the installed `generateFunctional` script.
+`uv` is not required — the tests invoke `python3` directly.
```
jmcarcell: Why is uv needed?

k4FWCore/helpers/tests/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+---
+
+## Setup
+
+```bash
+# 1. Clone and check out the branch
+git clone https://github.com/key4hep/k4FWCore.git
+cd k4FWCore
+git remote add ianna https://github.com/ianna/k4FWCore.git
```
jmcarcell: These instructions point to your remote

k4FWCore/CMakeLists.txt
```diff
@@ -18,6 +18,8 @@ limitations under the License.
 ]]
 
 gaudi_install(SCRIPTS)
+install(PROGRAMS helpers/generateFunctional
```
jmcarcell, line 21: None of the tests are run in CI, which means there is no guarantee that they work after a PR is merged. In general all the tests that are being added have a different setup from what is in the repo (individual scripts instead of commands that ctest can run). I would much prefer to be a single way of running the tests.

k4FWCore/helpers/tests/README.md
ianna: The tests invoke `python3` directly and do not use the script's shebang, so `uv` is not required.

k4FWCore/helpers/tests/README.md
ianna: Yes, because this is for testing this PR.

k4FWCore/CMakeLists.txt
ianna: Thanks @jmcarcell ! I've updated the PR. Here's what changed:

- **k4FWCore/helpers/tests/CMakeLists.txt (new)** — registers all 9 tests with `add_test`, passing `GENERATEFUNCTIONAL` so no install is needed

- **k4FWCore/CMakeLists.txt** — adds `add_subdirectory(helpers/tests)` under `BUILD_TESTING`

- **_test_common.sh** — checks `GENERATEFUNCTIONAL` env var first (set by CTest), falls back to PATH/source

- **tests/README.md** — updated to show `ctest -R GenerateFunctional` as the primary way to run

ianna:
@jmcarcell and @andresailer - please review. Thanks!

ianna:
All tests pass -- including the ones with generated by the script classes. The failing workflow is due to a needed change that has already been fixed in EDM4hep: https://github.com/key4hep/EDM4hep/pull/497 but that change is not in the latest tag/release. I'm confident the PR can me merged.

@jmcarcell and @andresailer - please review. Thanks!

k4FWCore/helpers/tests/test_runtime_transformer.sh
```diff
@@ -0,0 +1,31 @@
...
+## See the License for the specific language governing permissions and
+## limitations under the License.
+##
+
+# test_runtime_transformer.sh — build-test: k4FWCore Transformer with runtime outputs
+source "$(dirname "${BASH_SOURCE[0]}")/_test_common.sh"
+
+run_cmake_build MyRuntimeTransformer \
+    -i 'edm4hep::MCParticleCollection:InputCollections' \
+    --runtime-inputs 'edm4hep::MCParticleCollection:InputCollections:MCParticles' \
```
jmcarcell, line 26: The combination of these two does not let you to have both normal and vector inputs at the same time

k4FWCore/helpers/generateFunctional
```diff
@@ -0,0 +1,987 @@
+#!/usr/bin/env -S uv run --script
```
jmcarcell, line 1: Does this force to have uv available? Why?

k4FWCore/helpers/README.md
```diff
@@ -0,0 +1,284 @@
...
+## How to run it
+
+There are three equivalent ways to invoke the script:
+
+```bash
+# 1. Recommended — uv resolves Python and Jinja2 from the PEP 723 block.
+uv run generateFunctional MyProducer -o 'edm4hep::MCParticleCollection:MCParticles'
+
+# 2. Direct execution via the shebang (requires uv on PATH).
+chmod +x generateFunctional
```
jmcarcell, line 54: Why not make it an executable?

jmcarcell:
Given that there is a lot of AI-generated code here, have you tried running it? Do you think it is something that people will use instead of going to an LLM? Would an LLM use it instead of reading one of the existing ones? Would you use it over telling a LLM to write an algorithm? I have been playing with it, and you still have to type the types and names of the collections at a minimum. In addition, this is not very easy to maintain, the template is very complex and one (or a LLM) would have to look for where everything is defined.

ianna:
> Given that there is a lot of AI-generated code here, have you tried running it? Do you think it is something that people will use instead of going to an LLM? Would an LLM use it instead of reading one of the existing ones? Would you use it over telling a LLM to write an algorithm? I have been playing with it, and you still have to type the types and names of the collections at a minimum. In addition, this is not very easy to maintain, the template is very complex and one (or a LLM) would have to look for where everything is defined.

I’ve run it, and I see it mainly as a convenience tool. It gives you a clean, framework‑correct scaffold when you already know the types and collections, but it won’t replace asking an LLM to write an algorithm. And yes, the template is complex — that’s the cost of covering all the framework edge cases.

Where it does help is as input to an LLM: the generated scaffold gives the model a stable structure to refine instead of inventing everything from scratch.

Thanks for reviewing it.

k4FWCore/helpers/generateFunctional
ianna: No longer — the shebang is now `#!/usr/bin/env python3`, so it runs under whatever Python + Jinja2 the environment provides (which is how the tests invoke it and how the Key4hep stack ships Jinja2). `uv` isn't required anymore: the PEP 723 metadata block is kept so `uv run --script generateFunctional …` still works for anyone who wants an isolated environment, but it's optional. If Jinja2 isn't importable, the script now prints a clear "install jinja2 or use uv" message instead of a traceback.

k4FWCore/helpers/README.md
ianna: Done — it's now committed with the executable bit set (mode 100755), so `./generateFunctional …` works directly, and after `cmake --install` it's available on `PATH` in the Key4hep environment.

k4FWCore/helpers/tests/test_runtime_transformer.sh
ianna: Good catch, but that's a property of the test, not the generator. `test_runtime_transformer.sh` uses a single input with the same key for both `-i` and `--runtime-inputs`, so it collapses to all-runtime. Inputs are marked scalar/vector independently, so mixing works when the scalar and runtime inputs have distinct keys, e.g.:

```
generateFunctional MyMixedInputs \
  -i 'edm4hep::MCParticleCollection:Fixed' 'edm4hep::TrackCollection:RuntimeTracks' \
  --runtime-inputs 'edm4hep::TrackCollection:RuntimeTracks:Tracks0,Tracks1' \
  -o 'edm4hep::MCParticleCollection:Output'
```

generates `Transformer<edm4hep::MCParticleCollection(const edm4hep::MCParticleCollection&, const std::vector<const edm4hep::TrackCollection*>&)>` — a scalar `KeyValue` input alongside a variable-length `KeyValues` one. I've added `test_mixed_inputs.sh` covering exactly this case; it compiles against the current Key4hep stack.

PR #375, opened 2026-02-12
Do not check for an exact formatting of the job options

jmcarcell:
Formatting rules changed in Gaudi in https://gitlab.cern.ch/gaudi/Gaudi/-/merge_requests/1881. By using eval, the values are normalized, for example "3.14000" (str) will be 3.14 (float) and the test will pass.

BEGINRELEASENOTES
- Do not check for an exact formatting of the job options to fix test with a newer version of Gaudi

ENDRELEASENOTES

Raised by Marco Clemencic.


jmcarcell:
Merging to make a new CI pipeline in Gaudi that builds and tests k4FWCore pass.

jmcarcell:
Yes, it's backwards compatible, at least with the versions we have in the nightlies and release but since formatting hasn't changed in a long time probably with all recent versions.

PR #390, opened 2026-04-01
Make it posssible to set metadata also from Tools

tmadlener:
BEGINRELEASENOTES
- Generalize the `k4FWCore::putParameter` and `k4FWCore::getParameter` functions to also work with Gaudi components other than Algorithms (e.g. `AlgTool`s).
- Add slightly more debug output to these functions

ENDRELEASENOTES

- [x] Includes #389

PR #391, opened 2026-04-01
Add more utility functions for setting collection level parameters

### LINKED ISSUE #359: Possibility to easily create collection metadata handles with IOSvc
With the deprecated `PodioDataSvc` it was possible to pass a collection `DataHandle` to the `MetaDataHandle`, e.g. like so

https://github.com/key4hep/k4FWCore/blob/c79b7a8a158c72483dd5577a20177602a5ce753c/test/k4FWCoreTest/src/components/k4FWCoreTest_cellID_writer.h#L52-L56

this would then automatically ensure that this is stored with the proper naming convention into the *metadata* Frame eventually.

Doing the same thing with the IOSvc backed `MetaDataHandle` is not quite as straight forward as it involves the following

```cpp
struct Algorithm final : public /* SomeFunctional */ {
  // .... the rest

  k4FWCore::MetaDataHandle<std::string> m_cellIDEncodingHandle{"dummy-string", Gaudi::DataHandle::Reader};

  StatusCode initizlize() final {
    m_cellIDEncodingHandle = m_cellIDEncodingHandle = k4FWCore::MetaDataHandle<std::string>{
        podio::collMetadataParamName(outputLocations()[0], edm4hep::labels::CellIDEncoding), Gaudi::DataHandle::Writer};

    m_cellIDEncodingHandle.put("<some-encoding-string>");

    return StatusCode::SUCCESS;
  }
};
```

As can be seen this requires to
- initializing a `MetaDataHandle` as a member with some dummy settings (as the `MetaDataHandle` doesn't have a default constructor)
- resetting the `MetaDataHandle` to have the proper descriptor in e.g. `initialize`
  - which in turn requires to manually call a podio function and figure out which `outputLocation` (or `inputLocation` when reading) to use

IMO this should be much easier.  I am currently not entirely sure what the best way would be. Technically, one could forego the whole `MetaDataHandle` entirely in this case and simply do

```cpp
// ... as above
  
  StatusCode initialize() final {
    k4FWCore::putParameter(podio::collMetadataParamName(outputLocations()[0], edm4hep::labels::CellIDEncoding),
                           "<some-encoding-string>");
  }

// ... as above
```

However, also that still requires to have knowledge of the `podio::collMetadataParamName` function and the `edm4hep::labels::CellIDEncoding`.

I think we should have some utility that at least hides the `podio::collMetadataParamName`.

tmadlener:
BEGINRELEASENOTES
- Add `k4FWCore::putCollectionParameter` and `k4FWCore::getCollectionParameter` to facilitate setting and getting collection level parameters / metadata
- Add `k4FWCore::putCellIDEncoding` and `k4FWCore::getCellIDEncoding` to facilitate setting and getting CellID encoding strings
- Adapt some tests to use these instead of `MetaDataHandle`s

ENDRELEASENOTES

- [x] Includes #390 
- [x] Includes #389 

These simply wrap `(get|put)Parameter`, but call `podio::collMetadataParamName` (with `edm4hep::labels::CellIDEncoding`) under the hood first before passing things on. This is a follow up on [this discussion](https://github.com/key4hep/k4SimGeant4/pull/89#discussion_r2685382218).

@jmcarcell it looks like currently the only test for CellID encoding goes through a non-functional algorithm. I think we should add one for functional algorithms as well. Do you have a preference on whether
- I should add a completely new Functional algorithm for this, or
- I should re-use an existing one (e.g. the `ExampleFunctionalProducerMultiple`)?

### DIFF
--- a/k4FWCore/include/k4FWCore/MetadataUtils.h
+++ b/k4FWCore/include/k4FWCore/MetadataUtils.h
@@ -105,6 +105,27 @@ getParameter(const std::string& name) {
   return metadataSvc->get<T>(name);
 }
 
+/// @brief Put a metadata parameter associated with a collection into the metadata
+///
+/// Internally builds the correct parameter name from the collection name and
+/// parameter name and then stores the value into the metadata frame via the
+/// MetadataSvc
+///
+/// @param collName The name of the collection to which the parameter should be
+///        associated
+/// @param paramName The name of the parameter
+/// @param value The value of the parameter
+/// @param comp The Gaudi component (algorithm, tool) that is saving the
+///             parameter, typically "this"
+/// @tparam T The type of the parameter value
+/// @tparam GaudiComp The type of the component. This will be deduced in
+///                   pretty much all of the use cases
+template <typename T, typename GaudiComp = Gaudi::Algorithm>
+void putCollectionParameter(const std::string& collName, const std::string& paramName, const T& value,
+                            const GaudiComp* comp) {
+  putParameter(podio::collMetadataParamName(collName, paramName), value, comp);
+}
+
 /// @brief Put the CellID encoding string for a collection into the metadata
 ///
 /// Internally builds the correct parameter name and then stores the encoding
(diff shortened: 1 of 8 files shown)


jmcarcell:
I think reusing is fine; any test is fine since they don't check for metadata.

tmadlener:
Added two more test cases to the already existing tests that do metadata for Functional algorithms.

PR #396, opened 2026-04-15
Add support for CMAKE_UNITY_BUILD

jmcarcell:
This PR adds support for CMAKE_UNITY_BUILD (see https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html), which reduces compilation time significantly. Note that podio (and EDM4hep by extension) already has it (https://github.com/AIDASoft/podio/pull/746).
This is now supported in Gaudi (see https://gitlab.cern.ch/gaudi/Gaudi/-/merge_requests/1887).

BEGINRELEASENOTES
- Changes to make `CMAKE_UNITY_BUILD` work:
  - Rename a few aliases so they don't collide when combining `.cpp` files
  - Use `std::format` instead of `fmt::format` when possible (not always necessary for Unity builds to work)
  - Fix the name of an algorithm that was copied and pasted and kept the old name (now colliding)

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/components/UniqueIDGenSvc.cpp
+++ b/k4FWCore/components/UniqueIDGenSvc.cpp
@@ -18,9 +18,9 @@
  */
 #include "UniqueIDGenSvc.h"
 
-#include <fmt/core.h>
 
 #include <cstddef>
+#include <format>
 #include <stdexcept>
 #include <string>
 
@@ -62,7 +62,7 @@ size_t UniqueIDGenSvc::getUniqueID(event_num_t evt_num, run_num_t run_num, const
     if (!inserted) {
       const auto& [id_evt, id_run, id_name] = it->second;
       throw std::runtime_error(
-          fmt::format("Duplicate ID for event number, run number and algorithm name: {}, {}, \"{}\". "
+          std::format("Duplicate ID for event number, run number and algorithm name: {}, {}, \"{}\". "
                       "ID already assigned to: {}, {}, \"{}\"",
                       evt_num, run_num, name, id_evt, id_run, id_name));
     }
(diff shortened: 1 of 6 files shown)


tmadlener (commented):
For the formatting story, I would propose to do maybe create a `fmt_compat` header along the lines of

```cpp
#ifdef __cpp_lib_format
#include <format>
#else 
#include <fmt/format.h>
#endif

namespace k4FWCore {
namespace fmt_c { // naming up for debate
#ifdef __cpp_lib_format 
using format;
#else 
using format;
#endif 
}
}
```

One could then use this as `fmt_c::format` everywhere and a full migration to `std::format` could be done rather easily. It would make the duplicated calls that are currently there go away. Plus it would make it very easy to just use this for now instead of having to guess whether we want `std::format` or `fmt::format`.

We do something similar for RNTuple in podio: https://github.com/AIDASoft/podio/blob/6ef2f5b276bd95636e9b395f7fa5bfcc147e3344/include/podio/RNTupleReader.h#L20-L28

tmadlener:
Alternatively, if we are sure that all our builds have `<format>` already, we can just ditch `<fmt/format.h>` entirely.

jmcarcell:
Ok I think that should be all cases that easy to change to `std::format`.

tmadlener (approved):
dev3 still not picked up the lastest EDM4hep it seems?

andresailer:
> dev3 still not picked up the lastest EDM4hep it seems?

dev3 has EDM4hep 1.0, but do you mean the HEAD of the branch?

tmadlener:
Yes, for it to succeed we would need the HEAD of EDM4hep. I understood one of the comments of @jmcarcell (in another PR) such that dev3 also has EDM4hep HEAD, but apparently I misunderstood there.

jmcarcell:
No, my mistake, EDM4hep in dev3 is the last tagged version.

PR #397, opened 2026-04-16
Bump the version of k4FWCore for versioning the usage of new features

jmcarcell:
Like in fd8bf0f65e3273ec25043ac3111e9e529abd7163

BEGINRELEASENOTES
- Bump the version of k4FWCore for versioning the usage of new features

ENDRELEASENOTES


CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ CMAKE_MINIMUM_REQUIRED(VERSION 3.15)
 project(k4FWCore LANGUAGES CXX)
 
 set(${PROJECT_NAME}_VERSION_MAJOR 1)
-set(${PROJECT_NAME}_VERSION_MINOR 5)
+set(${PROJECT_NAME}_VERSION_MINOR 6)
 set(${PROJECT_NAME}_VERSION_PATCH 0)
```
tmadlener, line 25:
```suggestion
set(${PROJECT_NAME}_VERSION_PATCH 99)
```

To follow our usual convention of signalling that this is a nightly?

Or we could just make a tag? Looking at the open PRs it seems like now would be as good a time as any.

jmcarcell: Yes, we can make a tag

jmcarcell:
Tags needed to be able to update this one in: k4SimGeant4, k4Reco and k4GaudiPandora

PR #399, opened 2026-04-20
Add properties related to existing parameters to the MetadataSvc

### LINKED ISSUE #301: Metadata access order isn't declared and resolved
Currently the access to metadata isn't being declared and resolved. For instance a result of two algorithms accessing the same metadata during initialize, the first writing and the second reading the same parameter, depends on the order of their initialization which we don't control directly.
This might not be happening now but probably something worth being aware of.

jmcarcell:
BEGINRELEASENOTES
- Add two properties, `ThrowIfDuplicate` and `SkipIfSameValue`. This will change the behaviour of existing code since now it will throw if there is a duplicate by default (although there are not many cases where the MetadataSvc is being used).

ENDRELEASENOTES

See https://github.com/key4hep/k4FWCore/issues/301. There is no data race because the read and write operations are locked with the same mutex in podio.

Close https://github.com/key4hep/k4FWCore/issues/301.

### DIFF
--- a/k4FWCore/components/MetadataSvc.h
+++ b/k4FWCore/components/MetadataSvc.h
@@ -19,6 +19,7 @@
 #ifndef FWCORE_METADATASVC_H
 #define FWCORE_METADATASVC_H
 
+#include "Gaudi/Property.h"
 #include "GaudiKernel/IDataProviderSvc.h"
 #include "GaudiKernel/Service.h"
 
@@ -35,6 +36,9 @@ public:
   StatusCode initialize() override;
   StatusCode finalize() override;
 
+  bool throwIfDuplicate() const override { return m_throwIfDuplicate; }
+  bool skipIfSameValue() const override { return m_skipIfSameValue; }
+
 protected:
   SmartIF<IDataProviderSvc> m_dataSvc;
 
@@ -44,6 +48,13 @@ protected:
   podio::Frame* getFrame() override;
   void setFrame(podio::Frame frame) override;
   void throwIfRunning() const override;
+
+private:
+  Gaudi::Property<bool> m_throwIfDuplicate{this, "ThrowIfDuplicate", true,
+                                           "Throw an exception if a metadata parameter is already set"};
+  Gaudi::Property<bool> m_skipIfSameValue{
+      this, "SkipIfSameValue", true,
+      "Silently skip putting a metadata parameter if the existing value is the same as the new value"};
 };
 
 #endif
--- a/k4FWCore/components/MetadataSvc.h
+++ b/k4FWCore/components/MetadataSvc.h
@@ -53,7 +53,7 @@ private:
   Gaudi::Property<bool> m_throwIfDuplicate{this, "ThrowIfDuplicate", true,
                                            "Throw an exception if a metadata parameter is already set"};
   Gaudi::Property<bool> m_skipIfSameValue{
-      this, "SkipIfSameValue", true,
+      this, "SkipIfSameValue", false,
       "Silently skip putting a metadata parameter if the existing value is the same as the new value"};
 };
 
--- a/k4FWCore/include/k4FWCore/IMetadataSvc.h
+++ b/k4FWCore/include/k4FWCore/IMetadataSvc.h
@@ -33,6 +33,9 @@ public:
 
   virtual void setFrame(podio::Frame frame) = 0;
 
+  virtual bool throwIfDuplicate() const = 0;
+  virtual bool skipIfSameValue() const = 0;
+
   template <typename T>
   void put(const std::string& name, const T& obj) {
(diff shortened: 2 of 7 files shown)


tmadlener:
Would it be worth having this as an (optional) argument to `setParameter`? Such that there is at least a bit of control over what will be in the parameters? Something like `bool overWriteIfExists`? In case the old value would be interesting `putParameter` would need to get a `std::optional<T>` return value in which the old value could be returned instead of just silently dropping it.

(I should know this) On the podio side this will actually overwrite the value of the parameter, right?

jmcarcell:
> (I should know this) On the podio side this will actually overwrite the value of the parameter, right?

Overwrite, yes.

> Would it be worth having this as an (optional) argument to `setParameter`? Such that there is at least a bit of control over what will be in the parameters? Something like `bool overWriteIfExists`? In case the old value would be interesting `putParameter` would need to get a `std::optional<T>` return value in which the old value could be returned instead of just silently dropping it.

What about globally from the metadata service? With the approach above one would have to set that to false not to overwrite. There is another possibility, not letting overwriting the current parameter to other algorithms. Globally either allowing it or not sounds less complicated than per-parameter to me.
Why would one want the old value of the parameter? It's a value that is not going to appear anywhere.

tmadlener:
> What about globally from the metadata service? With the approach above one would have to set that to false not to overwrite.

Yes, that would also make sense, I think. The default should probably `false` to uncover potential issues of overwriting things.

> Why would one want the old value of the parameter? It's a value that is not going to appear anywhere.

There are a few potential use cases, I could imagine (not all of them make sense in the end probably)
- Simply checking whether the overwritten value is the same (this could potentially be internalized into the `MetadataSvc`). This would give back some control to the user to trigger or not trigger an error. It might be OK to set the same parameter multiple times if it's always the same value (e.g. CellIDEncodings might be set by multiple algorithms)
- "Updating" a parameter; it might be useful to e.g. set some initial value in `initialize` and then have an updated value in `finalize`. I don't think I particularly like this use case and we should probably find a better way to do this in case it is really needed.

jmcarcell:
Ok, so we agree on a global switch that throws by default in the metadata service and then an option not to complain if the value is the same. I prefer not adding getting the old value in `putParameter` since the use-case it's not clear and can always be changed later if needed, and since now it is void that means that change is always possible.

tmadlener:
Sounds reasonable to me and will automatically flush out all potentially problematic cases that are currently in place.

k4FWCore/components/MetadataSvc.h
```diff
@@ -43,6 +47,13 @@ class MetadataSvc : public extends<Service, IMetadataSvc> {
   const podio::Frame* getFrame() const override;
   podio::Frame* getFrame() override;
   void setFrame(podio::Frame frame) override;
+
+private:
+  Gaudi::Property<bool> m_throwIfDuplicate{this, "ThrowIfDuplicate", true,
+                                           "Throw an exception if a metadata parameter is already set"};
+  Gaudi::Property<bool> m_skipIfSameValue{
+      this, "SkipIfSameValue", true,
```
tmadlener:
```suggestion
      this, "SkipIfSameValue", false,
```
Either `true` or `false` will change existing behavior, because currently the last one to write with a given key wins. I would then go for `false` as default because that is at least a visible failure rather than a potentially silent error that will be somewhat painful to debug.

jmcarcell: Changed to false

tmadlener (approved):
I assume failing CI will pass tomorow when https://github.com/key4hep/EDM4hep/pull/497 is there.

jmcarcell:
CI is happy, dev3 builds the latest tag of EDM4hep.

jmcarcell:
I'm playing with adding parameters from Python to the Metadata (pre-fill) but I think this is needed if we want to check in the case where the parameters are already in a frame when they are read, any more opinions on this @tmadlener? I'd like to fix the Marlin wrapper (or we can revert not allowing setting parameters in the event loop).

tmadlener:
Since those will be properties with known values, they could be exempt from the overwrite checks. I think the existing `SkipIfSameValue` is already a good first version of this, because then one would simply turn that on when pre-setting some values. One would still get a failure if the value from the config and the value from the algorithm differ (unless `ThrowIfDuplicate` is set to `False).

It might be necessary to have more granular control over this, e.g. one behavior for parameters that are set from algorithms and another for ones that come from the `MetadataSvc` pre-setting, but I can't think of one right now, so I would go with what we have for now.

jmcarcell:
We have 3 places for parameters to come from, from the Frame if reading, from the MetadataSvc in python and set at `initialize()` or `finalize()` so this gets a bit complicated. Let's see in a coming PR, I think this is good as it is without worrying about setting parameters from Python.

PR #400, opened 2026-04-20
Disallow setting metadata parameters in the event loop

### LINKED ISSUE #302: Should writing metadata during event loop be allowed for MetadataSvc?
Putting parameters during event loop (algorithm execution) isn't allowed for the old `MetaDataHandle`:

https://github.com/key4hep/k4FWCore/blob/cc6a28a245592a5b54667925a4b2e3809e6eba54/k4FWCore/include/k4FWCore/MetaDataHandle.h#L127-L131

For the new `MetadataSvc` it's possible as this check seems to be missing. Is this an intended feature or an oversight?

jmcarcell:
Fix https://github.com/key4hep/k4FWCore/issues/302

BEGINRELEASENOTES
- Disallow setting metadata parameters in the event loop
- Add a test that checks that it is not possible

ENDRELEASENOTES

Create a function since this check is also needed in the `put` override for edm4hep::utils::ParticleIDMeta.

### DIFF
--- a/k4FWCore/include/k4FWCore/IMetadataSvc.h
+++ b/k4FWCore/include/k4FWCore/IMetadataSvc.h
@@ -35,6 +35,7 @@ public:
 
   template <typename T>
   void put(const std::string& name, const T& obj) {
+    throwIfRunning();
     getFrameForWrite()->putParameter(name, obj);
   }
 
@@ -50,6 +51,7 @@ public:
 protected:
   virtual podio::Frame* getFrame() = 0;
   virtual const podio::Frame* getFrame() const = 0;
+  virtual void throwIfRunning() const {}
 
 private:
   podio::Frame* getFrameForWrite() {
@@ -63,6 +65,7 @@ private:
 template <>
 inline void IMetadataSvc::put<edm4hep::utils::ParticleIDMeta>(const std::string& collName,
                                                               const edm4hep::utils::ParticleIDMeta& pidMetaInfo) {
+  throwIfRunning();
   edm4hep::utils::PIDHandler::setAlgoInfo(*getFrameForWrite(), collName, pidMetaInfo);
 }
 
--- a/k4FWCore/include/k4FWCore/IMetadataSvc.h
+++ b/k4FWCore/include/k4FWCore/IMetadataSvc.h
@@ -35,7 +35,6 @@ public:
 
   template <typename T>
   void put(const std::string& name, const T& obj) {
-    throwIfRunning();
     getFrameForWrite()->putParameter(name, obj);
   }
 
@@ -55,6 +54,7 @@ protected:
 
 private:
   podio::Frame* getFrameForWrite() {
+    throwIfRunning();
     if (!getFrame()) {
       setFrame(podio::Frame());
     }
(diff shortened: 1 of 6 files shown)


k4FWCore/include/k4FWCore/IMetadataSvc.h
```diff
@@ -35,6 +35,7 @@ class IMetadataSvc : virtual public IInterface {
 
   template <typename T>
   void put(const std::string& name, const T& obj) {
+    throwIfRunning();
     getFrameForWrite()->putParameter(name, obj);
   }
```
m-fila, line 39: Would it make sense to move `throwIfRunning` check into `getFrameForWrite`? So there is no need to put it by hand for the specializations

jmcarcell: Then one will want to call `getFrameForWrite` as soon as possible to check if it fails? Not evident, but copying and pasting `throwIfRunning` is also easy to forget. I'm not sure there will be many specializations.

tmadlener: Do we still need the `IMetadataSvc` to be available publicly? I think everyone should go through `(get|put)Parameter` in any case(?). What are the use-cases for having the interface?

edit: Just did a quick check with grepping through all key4hep repositories and the only use is in k4MarlinWrapper and that will be gone with https://github.com/key4hep/k4MarlinWrapper/pull/475

m-fila: > Then one will want to call getFrameForWrite as soon as possible to check if it fails? Not evident, but copying and pasting throwIfRunning is also easy to forget. I'm not sure there will be many specializations.

yeah, I mean probably hiding it deeper will make it harder to accidentally forget

k4FWCore/include/k4FWCore/IMetadataSvc.h
```diff
@@ -50,6 +51,7 @@ class IMetadataSvc : virtual public IInterface {
 protected:
   virtual podio::Frame* getFrame() = 0;
   virtual const podio::Frame* getFrame() const = 0;
+  virtual void throwIfRunning() const {}
```
m-fila, line 53: Is there a reason to make it virtual? From my understanding this should be always a check about Gaudi state machine which would be independend of how derived class manages its frame. Since the IMetedataSvc already isn't a pure interface class

jmcarcell: `targetFSMState()` and `name()` are not in `IInterface` so I can not implement it there?

m-fila: aaa right, nevermind then

k4FWCore/include/k4FWCore/IMetadataSvc.h
jmcarcell: `throwIfRunning()` moved as Mateusz proposed.

> Do we still need the `IMetadataSvc` to be available publicly? I think everyone should go through `(get|put)Parameter` in any case(?). What are the use-cases for having the interface?

Yes it's not needed for users. However, `SmartIF` wants an interface and that's used in IOSvc. I think this is the "standard" way. Could it be done without an interface?

tmadlener: I see, then also moving it out of the public headers is not easily done, I guess?

jmcarcell: No I don't think so, `MetadaUtils.h` needs it in a few places, like https://github.com/key4hep/k4FWCore/blob/83feba1ee53ed430b16e0b1297ac6c6125d79414/k4FWCore/include/k4FWCore/MetadataUtils.h#L87

jmcarcell:
I'll merge this later today if there aren't any more comments

tmadlener (approved):

m-fila (approved):

PR #403, opened 2026-04-22
Add a test using Tools

jmcarcell:
BEGINRELEASENOTES
- Add a test using Tools

ENDRELEASENOTES

Because we don't have any and they are used in several places. I tested if metadata was working for them, so may as well turn them into a test.


jmcarcell:
Wrong branch

PR #404, opened 2026-04-22
Add a test with Tools, since we do not have any

jmcarcell:
BEGINRELEASENOTES
- Add a test using Tools

ENDRELEASENOTES

Because we don't have any and they are used in several places. I tested if metadata was working for them, so may as well turn them into a test.


test/k4FWCoreTest/src/components/ExampleTool.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+  int getNumber() const override { return m_number; }
+
+private:
+  Gaudi::Property<int> m_number{this, "Number", 77, "The number returned by getNumber()"};
+};
+
+DECLARE_COMPONENT(ExampleToolC)
+
+struct ExampleToolUser final : k4FWCore::Consumer<void()> {
```
tmadlener: Mainly because I was a bit confused now here whether *User* is the (human) user or not, I would propose to rename to *ExampleToolUsingAlgorithm* (or maybe include *Functional* here to follow the usual naming convention that we have for the tests).

test/k4FWCoreTest/src/components/ExampleTool.cpp
```diff
@@ -0,0 +1,148 @@
...
+    if (auto sc = Algorithm::initialize(); sc.isFailure()) {
+      return sc;
+    }
+    if (auto sc = m_toolA.retrieve(); sc.isFailure()) {
+      return sc;
+    }
+    if (auto sc = m_toolB.retrieve(); sc.isFailure()) {
+      return sc;
+    }
+    return m_toolC.retrieve();
```
tmadlener:
```suggestion
    K4_GAUDI_CHECK(Algorithm::initialize());
    K4_GAUDI_CHECK(m_toolA.retrieve());
    K4_GAUDI_CHECK(m_toolB.retrieve());
    K4_GAUDI_CHECK(m_toolC.retrieve());
    return StatusCode::SUCCESS;
```

Mainly because people will look at this as an example and that would make them see `K4_GAUDI_CHECK`

tmadlener (commented):
Looks good. I have two minor comments below.

test/k4FWCoreTest/src/components/ExampleTool.cpp
jmcarcell: Renamed to `ExampleToolFunctional`.

test/k4FWCoreTest/src/components/ExampleToolFunctional.cpp
tmadlener, line 1: Should this be named `ExampleToolFunctional.cpp`? That would at least make it obvious that the algorithm is here, but then it also has some Tools, so I am not sure.

No strong preference from my side. Can also be merged as is.

jmcarcell: Renamed to `ExampleToolFunctional.cpp`.

PR #406, opened 2026-04-27
Allow pre-filling parameters for the MetadataSvc

jmcarcell:
BEGINRELEASENOTES
- Allow pre-filling parameters for the MetadataSvc, through int, float, double of str parameters
- Add a python function to allow passing a dictionary and letting Python infer which type is passed
- Modify a test checking that it is possible to pass parameters by hand or use the function to put them and also that overwriting happens and they are present in the resulting file

ENDRELEASENOTES

- Double or float by default?
- Global setting for setting all the parameters at `finalize()`? Otherwise it has to be a setting per-parameter
- Does this solve what we need for k4MarlinWrapper?

### DIFF
--- a/k4FWCore/components/MetadataSvc.cpp
+++ b/k4FWCore/components/MetadataSvc.cpp
@@ -40,8 +40,8 @@ StatusCode MetadataSvc::initialize() {
     error() << "Unable to locate the EventDataSvc" << endmsg;
     return StatusCode::FAILURE;
   }
-  bool hasParameters = !m_intParameters.empty() || !m_floatParameters.empty() ||
-                       !m_doubleParameters.empty() || !m_stringParameters.value().empty();
+  bool hasParameters = !m_intParameters.empty() || !m_floatParameters.empty() || !m_doubleParameters.empty() ||
+                       !m_stringParameters.value().empty();
   if (hasParameters && !m_setAtFinalize) {
     // If there is a reader, then the metadata frame will be set from the IOSvc
     // to what is read from the Reader
--- a/k4FWCore/components/MetadataSvc.cpp
+++ b/k4FWCore/components/MetadataSvc.cpp
@@ -87,7 +87,7 @@ void MetadataSvc::applyPropertyParameters() {
     m_frame->putParameter(key, val);
   }
   for (const auto& [key, val] : m_floatParameters) {
-    m_frame->putParameter(key, val);
+    m_frame->putParameter<float>(key, static_cast<float>(val));
   }
   for (const auto& [key, val] : m_doubleParameters) {
     m_frame->putParameter(key, val);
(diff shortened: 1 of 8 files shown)


k4FWCore/components/MetadataSvc.cpp
```diff
@@ -52,7 +62,39 @@ void MetadataSvc::throwIfRunning() const {
...
+podio::Frame* MetadataSvc::getFrame() {
+  if (m_setAtFinalize && !m_paramsApplied) {
+    if (!m_frame) {
+      m_frame = std::make_unique<podio::Frame>();
+    }
+    applyPropertyParameters();
+    m_paramsApplied = true;
+  }
+  return m_frame.get();
+}
```
tmadlener: This can potentially return a `nullptr`, I think. In case the first `if` is `false`, then we fall through to the end and return an uninitialized `Frame*` at the end. Or is that condition somehow impossible to get in usage?

k4FWCore/components/MetadataSvc.cpp
```diff
@@ -40,6 +40,16 @@ StatusCode MetadataSvc::initialize() {
...
   }
+  bool hasParameters = !m_intParameters.empty() || !m_floatParameters.empty() || !m_doubleParameters.empty() ||
+                       !m_stringParameters.value().empty();
+  if (hasParameters && !m_setAtFinalize) {
+    // If there is a reader, then the metadata frame will be set from the IOSvc
+    // to what is read from the Reader
+    if (!m_frame) {
+      m_frame = std::make_unique<podio::Frame>();
+    }
+    applyPropertyParameters();
```
tmadlener, line 49:
```suggestion
    applyPropertyParameters();
    m_paramsApplied = true;
```
I don't yet fully get the logic when this should be toggled, so this is more of a guess. Maybe this should go into `applyPropertyParameters`?

k4FWCore/components/MetadataSvc.cpp
```diff
@@ -40,6 +40,16 @@ StatusCode MetadataSvc::initialize() {
     error() << "Unable to locate the EventDataSvc" << endmsg;
     return StatusCode::FAILURE;
   }
+  bool hasParameters = !m_intParameters.empty() || !m_floatParameters.empty() || !m_doubleParameters.empty() ||
+                       !m_stringParameters.value().empty();
+  if (hasParameters && !m_setAtFinalize) {
```
tmadlener: Is the `hasParameters` check necessary? Other than potentially creating a `Frame` nothing would change if this check was not here. Additionally in `getFrame` we do very similar logic without this check.

tmadlener (commented):
I will give this a go later today, but this looks like it should fix the marlin wrapper problems.

k4FWCore/components/MetadataSvc.cpp
jmcarcell: This is now only a getter and what was happening here happens in the `Writer::finalize()`.

k4FWCore/components/MetadataSvc.cpp
jmcarcell: `m_paramsApplied` is now moved inside `applyPropertyParameters`.

k4FWCore/components/MetadataSvc.cpp
jmcarcell: Removed, should only change creating an empty `Frame`.

PR #408, opened 2026-04-28
Make OutputCollection of CollectionMerger and input and output collections of EfficiencyFilter take a single argument

tmadlener:
It doesn't need to take multiple collections

BEGINRELEASENOTES
- Make the `CollectionMerger.OutputCollection` accept a single collection (and disallow passing a list).
- Same for the `EfficiencyFilter`, for both its input and output collections.

ENDRELEASENOTES

Just stumbled over this. I suppose we never changed this after #345. Maybe because it would break in many places and we don't have an easy transparent migration for this?


jmcarcell:
Looks good, I removed all the use-cases in https://github.com/key4hep/k4MLJetTagger/pull/14 and https://github.com/key4hep/k4GaudiPandora/pull/29. In addition, I also applied the changes to tests and the `EfficiencyFilter`, that seems not to be used. These are not too annoying since the Python error is quite descriptive.

tmadlener:
@madbaron @samf25 this will likely break a few things for the muon collider.

PR #413, opened 2026-07-06
Update OverlayTiming to support BIB random mixing

madbaron:
This PR has been split: it now carries only the random background mixing. 
The TBB-based parallel reading of background frames moves to a follow-up PR, so that the two can be reviewed independently. It is based on #429, which simplifies the background particle index mapping and should go in first.

Beam-induced background is typically produced as a large number of files, each holding a single pseudo-event. Read as one sequential stream, such a group is walked in a fixed order, so a run overlaying fewer events than there are files
only ever sees the first few. `RandomMixBackgroundFiles = True` instead treats every file of a group as an independent event source and draws one per overlaid event, from a shuffled permutation that is reshuffled once exhausted. Entries of `BackgroundFileNames` may be directories, whose `.root` files are collected automatically, so a group of thousands of files does not have to be listed by hand. Each file keeps its own cursor and its event count is resolved only when it is first read, so the files are not all opened up front.

Since background particles usually dominate the output size, `MergeMCParticles =False` leaves the background `MCParticle` collection out entirely. Tracker hits then keep the momentum of the particle they came from instead of a relation to it, and calorimeter contributions get an empty particle.

Reads from a background group are now serialized with a mutex. `operator()` is const and Gaudi may run several event slots concurrently against one shared `EventHolder`, which was already unguarded on `main`.

### Notes for reviewers

Since this replaces the previous contents of the PR, the existing threads land as follows:
- The `EventHolder` locking question is answered by the mutex above; the worker thread and request queue this PR used to carry are gone with the TBB part.
- The comments on the background particle index maps are addressed in #429.
- Whether opening a reader per read is too costly still applies here, and remains deferred as agreed.
- The Gaudi scheduler and TBB `find_package` threads belong to the follow-up PR.

BEGINRELEASENOTES
- Add `RandomMixBackgroundFiles` to `OverlayTiming`, treating every file of a background group as an independent pseudo-event source and drawing one at random for each overlaid event, for background that is split across many single-event files
- Allow entries of `BackgroundFileNames` to be directories, whose `.root` files are collected automatically
- Add `MergeMCParticles` to `OverlayTiming`. With `false` the background MCParticles are not copied into the output: tracker hits keep the momentum of their originating particle instead of a particle relation, and calorimeter contributions get an empty particle
- Serialize the background file reads in `OverlayTiming`, which were unguarded against Gaudi processing several event slots concurrently
- Fail during `initialize()` with a message naming the group when a background group contains no `.root` files, instead of failing later or silently overlaying nothing
ENDRELEASENOTES


madbaron:
The downstream build failure doesn't seem related to the changes - but I can't retrigger it.

andread3vita:
Hi! I tried this implementation, and if I simply set `overlay.RandomMixBackgroundFiles = True`, the overlay works using just a single background file instead of 40:


```
from Gaudi.Configuration import INFO

from k4FWCore import ApplicationMgr
from k4FWCore import IOSvc
from Configurables import EventDataSvc
from Configurables import OverlayTiming
from Configurables import UniqueIDGenSvc

from pathlib import Path


background_base = Path("/eos/experiment/fcc/ee/simulation/key4hep_2026_04_20/91GeV/IDEA_o1_v03/IPC_Z_background")
background_file_list = []
for d in sorted(background_base.iterdir()):
    background_file_list.append(str(d))
            
id_service = UniqueIDGenSvc("UniqueIDGenSvc")
eds = EventDataSvc("EventDataSvc")
iosvc = IOSvc()
# iosvc.Input = "/afs/cern.ch/user/a/aloeschc/fcc_fullsim_testing_grounds/data/IDEA/IDEA_o1_v03/physics_events/p8_ee_Z_qqbar_ud/1000_91.188GeV_ISR_FSR/000/IDEA_o1_v03_1000_p8_ee_Z_qqbar_ud_91.188GeV_ISR_FSR.root"
iosvc.Input = "/afs/cern.ch/work/a/adevita/public/testBIB/IDEA_test/muonGunIDEAv3o1.root" 

iosvc.Output = "IDEA_o1_v03_OverlayIPC_test.root"

overlay = OverlayTiming()
overlay.MCParticles = "MCParticles"
overlay.BackgroundMCParticleCollectionName = "MCParticles"
overlay.SimTrackerHits = ["DCHCollection", "MuonSystemCollection", "SiWrDCollection", "SiWrBCollection", "VertexBarrelCollection", "VertexEndcapCollection", "PreshowerSystemCollection"]
overlay.SimCalorimeterHits = []
overlay.OutputSimTrackerHits = ["OverlayDCHCollection", "OverlayMuonSystemCollection", "OverlaySiWrDCollection", "OverlaySiWrBCollection", "OverlayVertexBarrelCollection", "OverlayVertexEndcapCollection", "OverlayPreshowerSystemCollection"]
overlay.OutputSimCalorimeterHits = []
overlay.OutputMCParticles = "OverlayMCParticles"
overlay.OutputCaloHitContributions = []
overlay.AllowReusingBackgroundFiles = True
overlay.CopyCellIDMetadata = True
overlay.NBunchtrain = 41          # total BX in train
overlay.NumberBackground = [1]    # one background event per BX
overlay.Delta_t = 20              # ns between BX
overlay.PhysicsBX = 21            # puts physics at 21 with 20 before & 30 after (allow for hits 200ns after event time)
overlay.Poisson_random_NOverlay = [False]
overlay.StartBackgroundEventIndex = -1
# overlay.BackgroundFileNames = [
#       background_file_list
# ]

overlay.RandomMixBackgroundFiles = True
overlay.BackgroundFileNames = [["/eos/experiment/fcc/ee/simulation/key4hep_2026_04_20/91GeV/IDEA_o1_v03/IPC_Z_background"]]

overlay.TimeWindows = {"MCParticles": [-400, 400], "DCHCollection": [-400, 400], "MuonSystemCollection": [-20, 0], "SiWrDCollection": [-20, 0],"SiWrBCollection": [-20, 0], "VertexBarrelCollection": [-20, 0],"VertexEndcapCollection": [-20, 0], "PreshowerSystemCollection": [-20, 0]}

iosvc.outputCommands = ["drop *", "keep OverlayDCHCollection*", "keep OverlaySiWrDCollection*", "keep OverlaySiWrBCollection*", "keep OverlayVertexBarrelCollection*", "keep OverlayVertexEndcapCollection*", "keep OverlayMC*", "keep *EventHeader*"]


ApplicationMgr(TopAlg=[overlay],
               EvtSel="NONE",
               EvtMax=1,
               ExtSvc=[eds],
               OutputLevel=INFO,
               )
```

ArinaPon:
Hello! while testing the implementation, I think I might have an idea why only one background file was being used across the bunch train.

`fileIndices` is shuffled before the BX loop, but `k` starts again from `0` for every BX. With `NumberBackground = [1]`, every BX therefore selects:

```cpp
fileIndices[0]
```

And I guess this means that the same randomly selected background file is reused for all  BXs of the signal event.

I tried adding a counter outside the BX loop:

```cpp
size_t fileCursor = 0;
```

and changing the file selection to:

```cpp
const int fileIndex =
    m_randomMix ? fileIndices[fileCursor++ % fileIndices.size()] : 0;
```

With this change, the code continues through the shuffled list instead of starting again from the first file for every BX.

I rebuilt and tested this with 120 IPC background files and the overlay completed successfully, the different events showed different background patterns.

madbaron:
Thanks for the checks @andread3vita and the suggestion for a fix @ArinaPon.
I ended up implementing it ~1:1, with the main difference that on wrap of the fileCursor I reshuffle so that there is no recurring pattern in overlaid background events.

doc/OverlayTiming.md
```diff
@@ -93,3 +96,41 @@ ApplicationMgr(
...
+```python
+overlay.RandomMixBackgroundFiles = True
+overlay.BackgroundFileNames = [["/path/to/bib_files/"]]
+```
+
+## Parallel background reading
+
+With many large background files the algorithm is dominated by reading and
+decompressing them. Set `OverlayThreads` to a value greater than 1 to read and
+decompress the background files of a single event on several threads:
```
tmadlener, line 119: Conceptually this might interfere with Gaudis internal scheduling (even if we also use tbb to do our multithreading). It's unclear to me whether the Gaudi internal tbb bits communicate with the tbb bits here.

There is precedent for doing this though as the CKF in k4ActsTracking also does some internal multithreading. This might need some policy discussion as it could imply different usage patterns for different community (e.g. run the general chain on a single thread but branch out to multi-threading in dedicated algorithms vs. running the full chain on multiple threads with Gaudi scheduling but no algorithm-internal multi-threading).

k4FWCore/components/OverlayTiming.h
```diff
@@ -44,35 +44,112 @@
...
+  // Cheap and I/O-free, so calling it serially (during the work-list build)
+  // does not limit read parallelism.
+  size_t reserve(int group, int file) {
+    std::lock_guard<std::mutex> lock(m_ioMutex);
+    size_t& entry = m_nextEntry[group][file];
+    const size_t e = entry;
+    const size_t total = m_totalNumberOfEvents[group][file];
+    entry = (total > 0) ? (entry + 1) % total : entry + 1; // wrap once the total is known
+    return e;
+  }
```
tmadlener, line 122: Do we need to lock this? The comment seems to imply it happens in a sequential piece of code(?).

k4FWCore/components/OverlayTiming.h
```diff
@@ -44,35 +44,112 @@
...
+
+  // Read a specific (group, file, rawEntry). Thread-safe: in random-mix mode
+  // each call opens its own reader and can run concurrently; in sequential mode
+  // the shared per-group reader is used under the mutex.
+  podio::Frame readAt(int group, int file, size_t rawEntry) {
+    if (m_randomMix) {
+      podio::Reader reader = podio::makeReader(m_fileNames[group][file]);
```
tmadlener, line 129: Intuitively this has a lot of overhead hat we are repeating for every event, because opening the file initializes a bunch of state in the reader which is quite expensive since the assumption is that this amortizes over many events and so we do a bunch of work upfront to save later in the event.

I can't say how much overhead this is in practice, but if there is a way to avoid it that might be a good idea.

k4FWCore/CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ gaudi_add_module(k4FWCorePlugins
                    components/Reader.cpp
                    components/UniqueIDGenSvc.cpp
                    components/Writer.cpp
-                 LINK Gaudi::GaudiKernel k4FWCore k4FWCore::k4Interface ROOT::Core ROOT::RIO ROOT::Tree EDM4HEP::edm4hep)
+                 LINK Gaudi::GaudiKernel k4FWCore k4FWCore::k4Interface ROOT::Core ROOT::RIO ROOT::Tree EDM4HEP::edm4hep TBB::tbb)
```
tmadlener, line 42: I suppose we pick up TBB transitively through Gaudi, because we do not explicitly `find_package` it. Not sure if we should make it explicit.

madbaron: The find_package sits in the repo-level CMakeLists.txt (https://github.com/madbaron/k4FWCore/blob/3b80ce75d171f3817da523c5494f08340ffb7029/CMakeLists.txt#L31)
Should it be placed here instead?

k4FWCore/components/OverlayTiming.h
madbaron: Within a single event this is only called from the sequential work-list build.
However, my understanding is that operator() is const and the scheduler may run several events concurrently that all share this EventHolder, so the lock is still needed.
(The code comment is misleading and I'll fix it)

k4FWCore/CMakeLists.txt
tmadlener: No, that is fine. I was just blind apparently.

k4FWCore/components/OverlayTiming.h
madbaron: I had measured it on some synthetic files. It's real, but small.

Timing breakdown for the one-pseudo-event-per-file pattern:

| per-file size      | open (`makeReader`+`getEntries`) | `readEvent` | decompress/materialise | crop+merge |
|--------------------|----------------------------------|-------------|------------------------|------------|
| ~40 MB (BIB-like)  | **1.8%** (≈5 ms/file)            | 52%         | 35%                    | 12%        |
| 55 KB              | **62%** (≈1.2 ms/file)           | 33%         | 3%                     | 1%         |

So the repeated open costs ~5 ms/file, ≈2% of the overlay time for the ~40 MB files we use.

doc/OverlayTiming.md
madbaron: Indeed. In the ideal world you might want to allow users to do a combination of both, if possible.
For now, especially in colliders that are computationally challenging per event, being able to use MT inside the same event is much more important than multi-threading over events, which can be done trivially in batch jobs anyway.

k4FWCore/components/OverlayTiming.h
tmadlener: Ok, I think now I understand. The lock is there for "Gaudi does the multithreading"-mode.

k4FWCore/components/OverlayTiming.h
tmadlener: Thanks a lot for testing and the numbers. IIUC, the driving factor here is the size of the event / collection that you read and not the file-size per se? Does the 55 KB file have a realistic number of elements? I am trying to understand whether we should document that this random mixing as it is done at the moment might incur some (quite significant) overhead if it is used on small overlay events.

doc/OverlayTiming.md
tmadlener: Maybe @jmcarcell knows if functional algorithms can already propagate that to the Gaudi scheduler somehow. Otherwise the potential interplay will for now just be another thing to document.

k4FWCore/components/OverlayTiming.h
madbaron: >  Does the 55 KB file have a realistic number of elements?

Not at all - I think O(10s) MB will be the smallest file size to be ever produced

tmadlener: Alright. Then from my side I would be happy with just defering that as a potential optimization to a future PR.

doc/OverlayTiming.md
madbaron: With the help of Claude, I had a look at the Gaudi sources. My understanding is that there's no way for a functional algorithm to declare or hand its internal parallelism to the scheduler. But the two TBB layers aren't independent either.

The good news is we can't oversubscribe the machine; the bad news is the scheduler counts algorithms in flight, not threads, so it keeps dispatching while we're fanned out and conversely with ThreadPoolSize=1 the pipeline only gets ~2 threads regardless of OverlayThreads (ntokens bounds in-flight items, not concurrency).

One thing I've added as a precaution: the pipeline now runs inside tbb::this_task_arena::isolate(). The reasoning is that while the calling thread blocks in parallel_pipeline, TBB may steal another AlgTask onto it, and AlgTask::operator() sets the thread-local EventContext and calls whiteboard()->selectStore(slot) without restoring either. While I haven't observed this, it costs nothing, so I'd rather keep it than rely on arena timing.

I've documented the OverlayThreads / ThreadPoolSize interplay in doc/OverlayTiming.md: they draw from one pool, OverlayThreads > 1 pays off when ThreadPoolSize is small, and raising both just repartitions the same threads.

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -120,6 +164,161 @@ StatusCode OverlayTiming::initialize() {
...
+    std::vector<edm4hep::CaloHitContributionCollection>& ocaloHitContribs) const {
+  const auto availableCollections = backgroundEvent.getAvailableCollections();
+
+  if (std::find(availableCollections.begin(), availableCollections.end(), m_MCParticleCollectionName) ==
+      availableCollections.end()) {
+    warning() << "Collection " << m_MCParticleCollectionName << " not found in background event" << endmsg;
+  }
+
+  // To fix the relations we will need to have a map from old to new particle index
+  std::map<int, int> oldToNewMap;
```
davidlange6: Something like this ought to be able to avoid all of these intermediate collections. [perhaps this is also a reserve function to use for oparticles?]

```
if (m_mergeMCParticles) {
    const auto& bgParticles     = backgroundEvent.get<edm4hep::MCParticleCollection>(m_MCParticleCollectionName);
    const int   offset          = static_cast<int>(oparticles.size());
    const int   n_bgParticles   = static_cast<int>(bgParticles.size());
  
    // Pass 1: clone all background particles into the output
    for (int i = 0; i < n_bgParticles; ++i) {
      auto npart = bgParticles[i].clone(false);
      npart.setTime(bgParticles[i].getTime() + timeOffset);
      npart.setOverlay(true);
      oparticles.push_back(npart);
    }
  
    // Pass 2: wire parent/daughter relations. New index = offset + old index.
    for (int i = 0; i < n_bgParticles; ++i) {
      for (const auto& parent : bgParticles[i].getParents()) {
        const int idx = parent.getObjectID().index;
        if (idx < 0 || idx >= n_bgParticles)
          continue;
        oparticles[offset + i].addToParents(oparticles[offset + idx]);
      }
      for (const auto& daughter : bgParticles[i].getDaughters()) {
        const int idx = daughter.getObjectID().index;
        if (idx < 0 || idx >= n_bgParticles)
          continue;
        oparticles[offset + i].addToDaughters(oparticles[offset + idx]);
      }
    }
  }

```

k4FWCore/components/OverlayTiming.h
davidlange6: i guess i had a similar question - this test was done from storage local to the cpu or with files on some distributed file system (eg, eos)?

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -120,6 +164,161 @@ StatusCode OverlayTiming::initialize() {
...
+
+      if (!((simTrackerHit.getTime() + timeOffset > this_start + tof) &&
+            (simTrackerHit.getTime() + timeOffset < this_stop + tof))) {
+        continue;
+      }
+      auto nhit = simTrackerHit.clone(false);
+      nhit.setOverlay(true);
+      nhit.setTime(simTrackerHit.getTime() + timeOffset);
+      if (m_mergeMCParticles) {
+        nhit.setParticle(oparticles[oldToNewMap[simTrackerHit.getParticle().getObjectID().index]]);
```
davidlange6: given my above comment, this would need to change to something like
```
 const int pidx = simTrackerHit.getParticle().getObjectID().index;
  if (pidx >= 0 && pidx < n_bgParticles)
    nhit.setParticle(oparticles[offset + pidx]);
```

k4FWCore/components/OverlayTiming.h
madbaron: This was with local storage

k4FWCore/components/OverlayTiming.cpp
madbaron: Thanks - happy to take this up after rebasing on main with https://github.com/key4hep/k4FWCore/pull/425

madbaron:
Now that the code is rebased on main, I'd split this off in three PRs to make it a little easier to digest:
PR1) based on main, will address @davidlange6's comments: https://github.com/key4hep/k4FWCore/pull/429
PR2) based on PR1 (*this PR*), introduces the "RandomMix" features to the synchronous algorithm
PR3) based on PR2, will introduce the threading

k4FWCore/components/OverlayTiming.cpp
madbaron: Done in https://github.com/key4hep/k4FWCore/pull/429, since the code predates this PR.

k4FWCore/components/OverlayTiming.cpp
madbaron: Also done in https://github.com/key4hep/k4FWCore/pull/429 via a small `overlaid_particle_index(index, offset, nBgParticles)` helper so the three hit and contribution call sites keep the shape https://github.com/key4hep/k4FWCore/pull/425 gave them.

madbaron:
This PR currently ships without tests for the new behaviour. 
Rather than decide unilaterally, since the shared fixtures involved touch a few unrelated tests, it seems worth agreeing on the shape first (@jmcarcell @tmadlener).

### Three tests

| Test | Covers |
| --- | --- |
| `OverlayTimingRandomMix` | `RandomMixBackgroundFiles`: directory expansion, several independent sources in a group, reshuffle-on-wrap |
| `OverlayTimingNoMCParticleMerge` | `MergeMCParticles = false`: background particles absent, hits keeping the momentum of their originating particle, contributions with an empty particle |
| `OverlayTimingMT` | Gaudi's own multithreading (6 event slots / 6 threads), exercising the mutex that serializes `EventHolder` reads |

`OverlayTimingMT`  would configure `HiveWhiteBoard` + `HiveSlimEventLoopMgr` + `AvalancheSchedulerSvc` to test several event slots sharing one `EventHolder`, not about any threading inside the algorithm.

Possible changes to shared test fixtures:
1. A non-zero momentum on `part1` in `ExampleFunctionalProducerMultiple`. Without it the `MergeMCParticles = false` assertion is vacuous: the hit's momentum is `(0, 0, 0)` whether or not it was copied. The x component has to stay 0, because `k4FWCoreTest_CheckExampleEventData` asserts on `particles[0].getMomentum().x` — `(0, 2, 3)` satisfies both. This file feeds nine tests through the `ProducerMultipleFile` fixture.
2. Assertions in `CheckOutputFiles.py`** for the two new output files.

tmadlener:
Sorry, didn't pay enough attention here apparently.

> PR2) based on PR1 (this PR), introduces the "RandomMix" features to the synchronous algorithm
PR3) based on PR2, will introduce the threading

I think that would be a sensible split for this, which would also make it a bit easier to pin-point potential issues in the future after they have been (squash)merged

Re, the tests:
I think the main things to check would be
- ensure that the grouping and sampling (of files) is done properly
- assignment / relating of MCParticles works as expected (this might already be covered in #429?). In case the MCParticle is dropped, the hit should retain the correct momentum, and all the information it can (this is probably what the second row of the table is about)
- reproducibility with a random seed. Not sure if this is currently a feature.
- Once multithreading is there exercising the mutex (row three) is quite an obvious check to also do

I think as a general comment if currently existing test setups need some adjustments to make the (to be added) new tests easier to do, we can do that. E.g. we don't have to keep all the currently existing assertions the same. Especially if it makes a test case easier to understand if just adjusting e.g. the momentum of a generated particle such that two cases are easier to disambiguate. Ideally that is done in a single (or a series of commits) where nothing else is done, such that they can potentially be lifted into a different PR if necessary.

madbaron:
> Re, the tests: I think the main things to check would be
> 
> * ensure that the grouping and sampling (of files) is done properly
> * assignment / relating of MCParticles works as expected (this might already be covered in [Simplify the background particle index mapping #429](https://github.com/key4hep/k4FWCore/pull/429)?). In case the MCParticle is dropped, the hit should retain the correct momentum, and all the information it can (this is probably what the second row of the table is about)
> * reproducibility with a random seed. Not sure if this is currently a feature.

Thanks! I've added tests along those lines:
 - Test input: a dedicated background fixture (CreateOverlayBackgroundFiles.py) writes groups of single-event files. 
 - Grouping and sampling (OverlayTimingRandomMix): each group only draws from its own files. Directories expand to their .root files only, and a group can mix a directory with an explicitly listed file. Every full pass over a group uses each file exactly once, and the list is reshuffled between passes. Hits and calo contributions point to the particles copied for their own draw, which extends the #429 relation checks to several draws per event.
- Reproducibility: the per-event seed comes from UniqueIDGenSvc. The same configuration is run again with the same Seed (it must give identical draws) and with a different one (it must differ). 
- MergeMCParticles = False (OverlayTimingNoMCParticleMerge): only signal particles are written. Background hits carry the momentum of their originating particle at creation and keep everything else, including the bunch-crossing time shift. Contributions keep PDG, energy, time and step position, with an empty particle.

The checks assert properties rather than exact sequences, since std::shuffle differs between libstdc++ and libc++. I also checked that each check fails when the corresponding piece of the algorithm is deliberately broken.

s6anloes:
Hi, I have a question about reproducability when files contain more than one event.
At the moment it is ture for IPC background that typically files contain exactly one event, but based on conversations with one of the MDI experts, this will likely not stay the case, and is already not true for SR background.
Do I understand correctly that in the current version the functionality for running with multiple events per file is there, but the results are not reproducable?
Maybe it's worth to check if this can be fixed at this stage already.

P.S.: I'll be also flagging two unrelated issues that Claude spotted when I was feeding it the code, but I have not verified that these are actual problems

k4FWCore/components/OverlayTiming.h
```diff
@@ -48,31 +48,97 @@
...
+
+    if (total == 0) {
+      throw GaudiException("No events found in background file " + m_fileNames[group][file], m_algName,
+                           StatusCode::FAILURE);
+    }
+    if (entry >= total && !m_allowReuse) {
+      throw GaudiException("No more events in background file", m_algName, StatusCode::FAILURE);
+    }
+    podio::Frame frame = reader->readEvent(entry % total);
+    entry = (entry + 1) % total;
```
s6anloes: The `entry>=total` check likely never fires, because in line 139, `entry` is stored to be wrapping around `total`. So right now the `m_allowReuse` flag is not working as intended?

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -236,19 +268,34 @@ retType OverlayTiming::operator()(const edm4hep::EventHeaderCollection& headers,
...
     if (m_startWithBackgroundEvent >= 0) {
       info() << "Skipping to event: " << m_startWithBackgroundEvent << endmsg;
-      for (auto& entry : m_bkgEvents->m_nextEntry) {
-        entry = m_startWithBackgroundEvent;
+      for (auto& group : m_bkgEvents->m_nextEntry) {
+        for (auto& entry : group) {
+          entry = m_startWithBackgroundEvent;
+        }
       }
     }
```
s6anloes: Claude was flagging that it seems that this skip is re-applied on every event. In the sequential reading mode (random mix off), this would mean that every event see the same background events?

k4FWCore/components/OverlayTiming.h
madbaron: Correct, this is a general bug! See https://github.com/key4hep/k4FWCore/pull/434

k4FWCore/components/OverlayTiming.cpp
madbaron: Same here https://github.com/key4hep/k4FWCore/pull/434

madbaron:
> P.S.: I'll be also flagging two unrelated issues that Claude spotted when I was feeding it the code, but I have not verified that these are actual problems

Thanks these are known but pre-existent, so I factored them in a separate fix PR (https://github.com/key4hep/k4FWCore/pull/434).

I'm looking at the situation in terms of reproducibility when mixing files with different entries. If it turns out not to be reproducible, I'll open an issue and work out a solution in a follow-up PR.

s6anloes:
> Thanks these are known but pre-existent, so I factored them in a separate fix PR (https://github.com/key4hep/k4FWCore/pull/434).

Sorry, I did not realise they were known!
What is the reason for not doing the fix straight-away here? This way we wouldn't merge a version with known issues. 

> I'm looking at the situation in terms of reproducibility when mixing files with different entries. If it turns out not to be reproducible, I'll open an issue and work out a solution in a follow-up PR.

Thanks!

madbaron:
> > Thanks these are known but pre-existent, so I factored them in a separate fix PR (#434).
> 
> Sorry, I did not realise they were known! What is the reason for not doing the fix straight-away here? This way we wouldn't merge a version with known issues.

The bug is on main, not a feature of this PR, so I think having a separate PR handles it more cleanly. If the fix gets merged before this PR (which I think might be the case?), I'll rebase to pick it up.

PR #417, opened 2026-07-21
Don't require `equality_comparable` in `putParameter`

ianna:
Fixes the build against recent EDM4hep, where `edm4hep::utils::ParticleIDMeta` is no longer `std::equality_comparable` (its `algoType` member became private, removing the defaulted `operator==`). `putParameter` required that concept only for the `*existing == value` comparison behind `skipIfSameValue()`.

This relaxes the template constraint and guards the comparison with `if constexpr (std::equality_comparable<T>)`. For non-comparable metadata types the "skip if identical value" shortcut is bypassed; all other behaviour is unchanged.

BEGINRELEASENOTES
- Fix compilation of `putParameter` with metadata types that are not `std::equality_comparable` (e.g. `edm4hep::utils::ParticleIDMeta` in recent EDM4hep). The `skipIfSameValue()` optimization is now only applied when the type supports `==`.

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/include/k4FWCore/MetadataUtils.h
+++ b/k4FWCore/include/k4FWCore/MetadataUtils.h
@@ -48,7 +48,7 @@ concept Streamable = requires(std::ostream& os, const T& v) {
 ///             parameter, typically "this"
 /// @tparam GaudiComp The type of the component. This will be deduced in
 ///                   pretty much all of the use cases
-template <std::equality_comparable T, typename GaudiComp>
+template <typename T, typename GaudiComp>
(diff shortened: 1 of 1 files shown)


ianna:
@jmcarcell and @andresailer - this should allow all tests pass for https://github.com/key4hep/k4FWCore/pull/372

jmcarcell:
That has already been fixed in EDM4hep: https://github.com/key4hep/EDM4hep/pull/497 but that change is not in the latest tag/release. The change here would silently ignore anything that doesn't have a `operator==` defined instead of failing at compile-time.

PR #421, opened 2026-08-21
Update dev3 build workflow

jmcarcell:
BEGINRELEASENOTES
 - Update dev3 CI to use the shared Key4hep build workflow and action

ENDRELEASENOTES


jmcarcell:
Needs https://github.com/AIDASoft/podio/pull/993 to be in dev3 to pass

PR #423, opened 2026-09-01
Set the CellID encoding of overlaid collections in initialize()

madbaron:
This PR is a follow-up to the discussion in https://github.com/key4hep/k4GaudiPandora/pull/48

The collections produced by `OverlayTiming` only come into existence at runtime, so they carry no CellID encoding of their own. `CopyCellIDMetadata` takes care of that by copying the encoding from the corresponding input collection, but it did so in `finalize()`: the encoding reached the output file, while remaining invisible for the whole run. Downstream algorithms that need it like digitisers building a decoder once in their own `initialize()` therefore never saw it, and overlaid collections could not be digitised.

BEGINRELEASENOTES
- Move the `CopyCellIDMetadata` handling in `OverlayTiming` from `finalize()` to `initialize()`, so the CellID encoding of the overlaid collections is visible to downstream algorithms during the run and not only in the output file
- Remove `OverlayTiming::finalize()`, which no longer does anything
- Extend the `OverlayTiming` test to cover `CopyCellIDMetadata`, which was previously untested, and add `ExampleCellIDEncodingInitConsumer`, which looks up the CellID encoding of its input collection in `initialize()` and fails if it is not there

ENDRELEASENOTES


k4FWCore/components/OverlayTiming.cpp
```diff
@@ -101,6 +101,24 @@ StatusCode OverlayTiming::initialize() {
...
   }
 
+  // Copy the cellID encoding of the input collections over to the overlaid output collections.
+  if (m_copyCellIDMetadata) {
+    for (const auto& [input, output] :
+         {std::make_pair(inputLocations("SimTrackerHits"), outputLocations("OutputSimTrackerHits")),
+          std::make_pair(inputLocations("SimCalorimeterHits"), outputLocations("OutputSimCalorimeterHits"))}) {
+      for (size_t i = 0; i < input.size(); ++i) {
+        const auto value = k4FWCore::getParameter<std::string>(
+            podio::collMetadataParamName(input[i], edm4hep::labels::CellIDEncoding), this);
```
tmadlener:
```suggestion
        const auto value = k4FWCore::getCellIDEncoding(input[i], this);
```
is a thing since #391

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -101,6 +101,24 @@ StatusCode OverlayTiming::initialize() {
...
+        if (value.has_value()) {
+          k4FWCore::putParameter(podio::collMetadataParamName(output[i], edm4hep::labels::CellIDEncoding),
+                                 value.value(), this);
```
tmadlener:
```suggestion
          k4FWCore::putCellIDEncoding(output[i], value.value(), this);
```
see above

test/k4FWCoreTest/scripts/CheckOutputFiles.py
```diff
@@ -365,6 +365,16 @@ def check_metadata(filename, expected_metadata):
...
+# The encoding is copied from the input collection by CopyCellIDMetadata and,
+# because OverlayTiming does that in initialize(), it is also visible to
+# downstream algorithms during the run, not just in this file
+check_metadata(
+    "overlay_output.root",
+    {
+        "OverlaySimTrackerHits__CellIDEncoding": "system:5,side:-2,layer:9,module:8,sensor:8",
+    },
+)
+
```
tmadlener: I don't think we need this check here. We know (from other checks) that metadata ends up properly in the output file if things go through utilities and the Functional that you have written to test whether `OverlayTiming` puts things in early enough does the exact same check, just in memory.

test/k4FWCoreTest/CMakeLists.txt
tmadlener, line 1: Just for confirmation: We don't need to add a new test to the harness here, because we modify an existing one, right?

tmadlener (commented):
A comment and a question below. I think looks good as it is otherwise.

test/k4FWCoreTest/scripts/CheckOutputFiles.py
madbaron: Removed

test/k4FWCoreTest/CMakeLists.txt
madbaron: Correct, we just attach ExampleCellIDEncodingInitConsumer into TestOverlayTiming.py

tmadlener (approved):

test/k4FWCoreTest/src/components/ExampleCellIDEncodingInitConsumer.cpp
```diff
@@ -0,0 +1,65 @@
...
+/// initialize() and fails if it is not there.
+///
+/// This mirrors what real digitisers do: they need the encoding to build a
+/// decoder once, at initialization time, rather than for every event. It exists
+/// so that algorithms producing collections at runtime can be tested for
+/// publishing their cellID encoding early enough for such consumers to see it.
+struct ExampleCellIDEncodingInitConsumer final
+    : k4FWCore::Consumer<void(const edm4hep::SimTrackerHitCollection& input)> {
+  ExampleCellIDEncodingInitConsumer(const std::string& name, ISvcLocator* svcLoc)
+      : Consumer(name, svcLoc, KeyValues("InputCollection", {"SimTrackerHits"})) {}
```
jmcarcell:
```suggestion
      : Consumer(name, svcLoc, KeyValue("InputCollection", "SimTrackerHits")) {}
```

test/k4FWCoreTest/options/TestOverlayTiming.py
```diff
@@ -46,9 +64,16 @@
 overlay.Poisson_random_NOverlay = [False]
 overlay.NBunchtrain = 1
 overlay.TimeWindows = {"SimTrackerHits": [-10000, 10000]}
+overlay.CopyCellIDMetadata = True
+
+# Fails to initialize if OverlayTiming has not published the encoding of
+# OverlaySimTrackerHits by the time this algorithm is initialized
+encoding_consumer = ExampleCellIDEncodingInitConsumer("EncodingConsumer")
+encoding_consumer.InputCollection = ["OverlaySimTrackerHits"]
```
jmcarcell:
```suggestion
encoding_consumer.InputCollection = "OverlaySimTrackerHits"
```

PR #425, opened 2026-09-01
fix: protect against out of bounds access

madbaron:
This PR fixes a minor issue discovered in OverlayTiming when testing https://github.com/key4hep/k4FWCore/pull/413 at scale.
It would be good to merge this before https://github.com/key4hep/k4FWCore/pull/413, so that 413 would rebase cleanly

BEGINRELEASENOTES
  - Fix `OverlayTiming` indexing its output MCParticle collection with the `-1` of an unset relation. Calorimeter contributions without an attached particle caused an out-of-bounds access that crashed when the contributions were written out.
  - Background hits and contributions whose particle is not part of the copied background particles now keep their relation unset instead of being silently attached to the first particle of the background event.

ENDRELEASENOTES


k4FWCore/components/OverlayTiming.cpp
```diff
@@ -143,10 +155,14 @@ retType OverlayTiming::operator()(const edm4hep::EventHeaderCollection& headers,
   // Fix relations to point to the new particles
   for (size_t i = 0; i < particles.size(); ++i) {
     for (const auto& parent : particles[i].getParents()) {
-      oparticles[i].addToParents(oparticles[parent.getObjectID().index]);
+      if (const auto index = parent.getObjectID().index; index != -1) {
```
jmcarcell: When does the `-1` happen here?

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -188,7 +204,11 @@ retType OverlayTiming::operator()(const edm4hep::EventHeaderCollection& headers,
         within_time_window = true;
         // TODO: Make sure a contribution is not added twice
         auto newContrib = contrib.clone(false);
-        newContrib.setParticle(oparticles[contrib.getParticle().getObjectID().index]);
+        // The contribution may have no particle attached, in which case the
+        // relation is left unset rather than indexed with -1.
+        if (const auto index = contrib.getParticle().getObjectID().index; index != -1) {
```
jmcarcell: Correct

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -363,7 +388,10 @@ retType OverlayTiming::operator()(const edm4hep::EventHeaderCollection& headers,
                   add = true;
                   // TODO: Make sure a contribution is not added twice
                   auto newContrib = contrib.clone(false);
-                  newContrib.setParticle(oparticles[oldToNewMap[contrib.getParticle().getObjectID().index]]);
+                  if (const auto index = mapped_particle_index(oldToNewMap, contrib.getParticle().getObjectID().index);
```
jmcarcell, line 389: Ok

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -382,7 +410,10 @@ retType OverlayTiming::operator()(const edm4hep::EventHeaderCollection& headers,
                 if ((contrib.getTime() + timeOffset > this_start) && (contrib.getTime() + timeOffset < this_stop)) {
                   // TODO: Make sure a contribution is not added twice
                   auto newContrib = contrib.clone(false);
-                  newContrib.setParticle(oparticles[oldToNewMap[contrib.getParticle().getObjectID().index]]);
+                  if (const auto index = mapped_particle_index(oldToNewMap, contrib.getParticle().getObjectID().index);
```
jmcarcell, line 411: Ok

k4FWCore/components/OverlayTiming.cpp
madbaron: Not in a standard self-contained MCParticle collection.
However, I think that background overlay files where the millions of MCParticles are often trimmed or dropped entirely to keep file sizes manageable, need this guard.

jmcarcell:
I think it would be good to change all the comparisons to test `< 0` since there is also `invalid = -2` (https://github.com/AIDASoft/podio/blob/master/include/podio/ObjectID.h#L22) that will cause exactly the same problem when the comparison fails with `!= 1`. 
I think changing all the accesses to `.at` instead of using the brackets is also not a bad idea, I don't think performance will be affected much and it would have caught these issues instead of creating another element in the maps.

jmcarcell (commented):
These comments are for myself of which parts I am convinced that are a fix.

k4FWCore/components/OverlayTiming.cpp
jmcarcell: But the background ones are read from files so it's impossible for the parents or daughters not to point to something.

tmadlener: Depending on how you prepare the background files, I think it's possible to have files with parent / daughter relations that point to nothing. You can get into a similar state with a normal file, if you simply filter out e.g. the beam particles or anything in the MC-tree really. I would have to check to make sure, but in that case, I think podio will just start filling `-1` when persisting the relations.

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -43,9 +43,24 @@ inline float time_of_flight(const T& pos) {
...
+    return -1;
+  }
+  const auto it = oldToNewMap.find(oldIndex);
+  return it == oldToNewMap.end() ? -1 : it->second;
+}
+
 std::pair<float, float> OverlayTiming::define_time_windows(const std::string& collection_name) const {
   try {
-    return {m_timeWindows.value().at(collection_name)[0], m_timeWindows.value().at(collection_name)[1]};
+    return {m_timeWindows.value().at(collection_name).at(0), m_timeWindows.value().at(collection_name).at(1)};
```
jmcarcell: All brackets where changed to `at`, this is not needed

k4FWCore/components/OverlayTiming.cpp
jmcarcell: Ok, I did a test and indeed writing persists the relations with -1 so OK.

k4FWCore/components/OverlayTiming.cpp
madbaron: I think I've undone the relevant parts of my over-eager replace all (I was unsure of how to interpret your message).

If there's anything left, or anything that you'd like to move to at(), can you please mark it directly?

madbaron:
@tmadlener @jmcarcell clang-tidy started to fail on things that seem unrelated to this PR. What do we do?

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -234,14 +258,14 @@ retType OverlayTiming::operator()(const edm4hep::EventHeaderCollection& headers,
...
 
       int NOverlay_to_this_BX = 0;
 
-      if (m_Poisson[groupIndex]) {
-        NOverlay_to_this_BX = std::poisson_distribution<>(m_Noverlay[groupIndex])(rng_engine);
+      if (m_Poisson.value()[groupIndex]) {
+        NOverlay_to_this_BX = std::poisson_distribution<>(m_Noverlay.value()[groupIndex])(rng_engine);
       } else {
-        NOverlay_to_this_BX = m_Noverlay[groupIndex];
+        NOverlay_to_this_BX = m_Noverlay.value()[groupIndex];
```
jmcarcell:
```suggestion
      const int BX_number_in_train = permutation.at(bxInTrain);

int NOverlay_to_this_BX = 0;

      if (m_Poisson.value()[groupIndex]) {
        NOverlay_to_this_BX = std::poisson_distribution<>(m_Noverlay[groupIndex])(rng_engine);
} else {
        NOverlay_to_this_BX = m_Noverlay[groupIndex];
```
Unrelated

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -275,15 +299,15 @@ retType OverlayTiming::operator()(const edm4hep::EventHeaderCollection& headers,
...
-          npart.setTime(bgParticles[i].getTime() + timeOffset);
+          npart.setTime(bgParticles.at(i).getTime() + timeOffset);
           npart.setOverlay(true);
           oparticles.push_back(npart);
-          for (const auto& parent : bgParticles[i].getParents()) {
+          for (const auto& parent : bgParticles.at(i).getParents()) {
             parentDaughterMap[j].first.push_back(parent.getObjectID().index);
           }
-          for (const auto& daughter : bgParticles[i].getDaughters()) {
+          for (const auto& daughter : bgParticles.at(i).getDaughters()) {
```
jmcarcell:
```suggestion
          auto npart = bgParticles.[i]].clone(false);

          npart.setTime(bgParticles.[i].getTime() + timeOffset);
npart.setOverlay(true);
oparticles.push_back(npart);
          for (const auto& parent : bgParticles.[i].getParents()) {
parentDaughterMap[j].first.push_back(parent.getObjectID().index);
}
          for (const auto& daughter : bgParticles.[i].getDaughters()) {
```
Not needed

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -43,6 +43,21 @@ inline float time_of_flight(const T& pos) {
   return std::sqrt((pos[0] * pos[0]) + (pos[1] * pos[1]) + (pos[2] * pos[2])) / TMath::C() * 1e6;
 }
 
+// Index of the copied background particle a relation should point at, or -1 when
+// the relation has to be left unset. An unset relation is any negative index, both
+// podio::ObjectID::untracked (-1) and podio::ObjectID::invalid (-2), which is why
+// oldIndex is tested against 0 rather than against -1. std::map::operator[] cannot
+// be used for the lookup either: it default-constructs a 0 for a missing key, which
+// would silently attach the hit to the first particle of the background event
+// instead of leaving the relation unset.
```
jmcarcell: Tune down

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -188,14 +208,18 @@ retType OverlayTiming::operator()(const edm4hep::EventHeaderCollection& headers,
         within_time_window = true;
         // TODO: Make sure a contribution is not added twice
         auto newContrib = contrib.clone(false);
-        newContrib.setParticle(oparticles[contrib.getParticle().getObjectID().index]);
+        // The contribution may have no particle attached, in which case the
+        // relation is left unset rather than indexed with a negative index.
```
jmcarcell:
```suggestion
```
Obvious

test/k4FWCoreTest/options/TestAlgorithmWithTFile.py
```diff
@@ -21,7 +21,6 @@
 from k4FWCore import ApplicationMgr, IOSvc
 from Configurables import k4FWCoreTest_AlgorithmWithTFile, EventDataSvc
 
-
```
jmcarcell, line 24: Bring back the unrelated lines, forget about pre-commit for now

test/k4FWCoreTest/options/TestOverlayTimingNoParticles.py
```diff
@@ -0,0 +1,65 @@
+#
```
jmcarcell: Use the existing `TestOverlayTiming.py`, the file produced above can be passed as another background file. Maybe the names of the collections need to be adjusted.

k4FWCore/components/OverlayTiming.cpp
madbaron: I assume you meant shorten - let me know if you did not.

test/k4FWCoreTest/options/TestAlgorithmWithTFile.py
madbaron: I'm not sure what this comment is about :S

madbaron: @tmadlener explained me this was meant to be "line". Added back!

test/k4FWCoreTest/options/TestOverlayTimingNoParticles.py
madbaron: Done, and it turned out even smaller.

`functional_producer_multiple.root` is already a file whose `SimTrackerHits` have no particle set.
What was missing was the calorimeter side.  The producer now makes the signal ones in memory as a `TopAlg` of the test.

jmcarcell:
Reorganized the OverlayTiming test collections: the normal simulated tracker and calorimeter collections carry MCParticle relations, while dedicated WithoutParticleRelations collections exercise unset relations. The OverlayTiming fix itself is unchanged; the unrelated m_Poisson.value() access was removed.

Issue #332, opened 2025-06-25
NTuple service example


tmadlener:
We have some examples here that use the histogram service. Do we have similar examples (or can we come up with examples) that let us produce ntuples (outside of EDM4hep)? Would make some monitoring tasks a bit easier potentially. At the moment one would have to create a `podio::UserDataCollection` for every branch that we want to have in the ntuple which again requires changing the algorithm type in some cases.

PR #427, opened 2026-09-03
Refactor test components to functional algorithms

jmcarcell:
BEGINRELEASENOTES
 - Refactor test components to use functional algorithms.
   - Replace the legacy ROOT-file producer with a functional producer.
   - Replace the legacy Marlin-wrapper producer with a functional generic-collection producer.
   - Consolidate CellID metadata validation in the functional metadata consumer.
   - Centralize expected output collection definitions in the output-file check.

ENDRELEASENOTES


test/k4FWCoreTest/src/components/ExampleFunctionalProducerWithTFile.cpp
tmadlener, line 1: I think this test has been there mainly for historical reasons. I am not sure whether we should even keep this as an example at the moment. This version is not thread safe (but appears to be due to it being a Functional algorithm) and people will just copy this and then wonder why the output looks strange. 

I suppose the main use case for this would be something that would solve https://github.com/key4hep/k4FWCore/issues/332 (or the PR related to that https://github.com/key4hep/k4FWCore/pull/364)

test/k4FWCoreTest/src/components/ExampleFunctionalProducerWithMarlinWrapper.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "edm4hep/MCParticleCollection.h"
+#include "podio/CollectionBase.h"
+
+#include "k4FWCore/Producer.h"
+
+#include <string>
+
+// Produce a collection through a CollectionBase pointer, as the Marlin wrapper
+// does when it registers a collection in the event store.
+struct ExampleFunctionalProducerWithMarlinWrapper final : k4FWCore::Producer<podio::CollectionBase*()> {
```
tmadlener, line 29: I was not aware this was possible... Nice.

test/k4FWCoreTest/src/components/ExampleFunctionalMetadataConsumer.cpp
```diff
@@ -125,6 +142,11 @@ struct ExampleFunctionalMetadataConsumer final : k4FWCore::Consumer<void(const e
   }
 
 private:
+  Gaudi::Property<bool> m_checkMetadata{this, "CheckMetadata", true, "Check the standard metadata parameters"};
+  Gaudi::Property<std::string> m_cellIDCollection{this, "CellIDCollection", "",
+                                                  "Collection whose cellID encoding is checked"};
+  Gaudi::Property<std::string> m_expectedCellIDEncoding{this, "ExpectedCellIDEncoding", "",
+                                                        "Expected cellID encoding for CellIDCollection"};
```
tmadlener, line 149: At some point we should consider splitting this into smaller algorithms again that do only one thing, otherwise debugging tests becomes its own challenge again, and ideally you immediately see whats wrong from a failing test without too much thinking.

test/k4FWCoreTest/src/components/ExampleFunctionalProducerWithMarlinWrapper.cpp
jmcarcell: Also used here https://github.com/key4hep/k4FWCore/blob/main/k4FWCore/components/EfficiencyFilter.cpp#L37 and maybe more places

test/k4FWCoreTest/src/components/ExampleFunctionalMetadataConsumer.cpp
jmcarcell: On the other side having a different algorithm for everything means they all have to be compiled and maintained. At least the ones that produce collections or files should try to produce for as many other tests as possible. For this one I find the theme "test metadata" fits the test. Balance will have to be found.

test/k4FWCoreTest/src/components/ExampleFunctionalProducerWithTFile.cpp
jmcarcell: Added `isReEntrant()`. I think we have many algorithms that use TFile or TTree, I guess there is some value to make sure that it works.

tmadlener: Thanks. I think it's OK like this. At least people can copy from a complete example that serializes properly.

tmadlener (approved):

PR #429, opened 2026-09-09
Simplify the background particle index mapping

madbaron:
Follow-up to @davidlange6's review comment in https://github.com/key4hep/k4FWCore/pull/413#discussion_r3933115912. The code it points at predates that PR, so it is split out here against `main`.

`OverlayTiming` built two `std::map`s per background frame to translate background particle indices into output ones. They are redundant: the background particles are cloned in order, so the copy of background particle `i` always lands at `offset + i`.

Dropping them also fixes a guard on the relation wiring: it skipped a relation when the target particle had no relations of its own, rather than when the index pointed outside the background collection. A particle referenced as a daughter whose own parent list is empty, was therefore silently dropped.

BEGINRELEASENOTES
- Simplify the background particle merging in `OverlayTiming` by computing the index of a copied particle as `offset + i`, instead of building an old-to-new index map and a parent/daughter staging map for every background frame
- Fix `OverlayTiming` dropping relations between background particles when the referenced particle had no parents or daughters of its own. Whether a relation is kept now depends on its index pointing inside the background collection
- Extend the `OverlayTiming` test with a one-sided MCParticle relation and check the parent/daughter topology of the overlaid particles, which was previously untested
ENDRELEASENOTES


k4FWCore/components/OverlayTiming.cpp
```diff
@@ -43,15 +43,11 @@ inline float time_of_flight(const T& pos) {
...
-  if (oldIndex < 0) {
-    return -1;
-  }
-  const auto it = oldToNewMap.find(oldIndex);
-  return it == oldToNewMap.end() ? -1 : it->second;
+// Index of the copied background particle a relation should point at.
+// An index outside the background collection leaves the relation unset, which
+// covers both podio::ObjectID::untracked (-1) and podio::ObjectID::invalid (-2).
+inline int overlaid_particle_index(int oldIndex, int offset, int nBgParticles) {
+  return (oldIndex >= 0 && oldIndex < nBgParticles) ? offset + oldIndex : -1;
```
tmadlener, line 52: I am wondering if we ever run into problems here with overflow in the `offset + oldIndex` addition. We could in principle fix that with just using `size_t` everywhere. However, since podio collections are effectively also limited to the positive values of `int` for their sizes at the moment we would just be hiding potential issues. Maybe there should be a check (or at least an `assert`) here?

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -286,45 +282,33 @@ retType OverlayTiming::operator()(const edm4hep::EventHeaderCollection& headers,
...
-              // warning() << "Parent " << parent << " not found in background event" << endmsg;
-              continue;
+
+        // The relations can only be wired up once every particle has been copied.
+        // Relations pointing outside the background collection are left unset.
+        for (int i = 0; i < nBgParticles; ++i) {
+          for (const auto& parent : bgParticles[i].getParents()) {
+            if (const auto index = overlaid_particle_index(parent.getObjectID().index, offset, nBgParticles);
+                index >= 0) {
+              oparticles.at(offset + i).addToParents(oparticles.at(index));
```
tmadlener, line 310: `oparticles.at(offset + i)` is guaranteed to exist by construction, so we could drop the built-in range check in `at` and just use `oparticles[offset + i]`. The main difference is that the current version gives us an exception that we let escape to Gaudi where as `operator[]` will just crash. Performance-wise, I would assume branch prediction will probably do its thing and it's probably immeasurable in all the other work that is done here.

test/k4FWCoreTest/scripts/CheckOutputFiles.py
```diff
@@ -419,6 +428,22 @@ def check_metadata(filename, expected_metadata):
     if particle_overlay_flags[n_signal_mc:] != [True] * n_background_mc:
         raise RuntimeError("Background particles should be flagged as overlay")
 
+    # The producer makes part2 a daughter of part1 without filling in part2's own
+    # parent list. Both blocks, signal and overlaid background, have to come out
+    # with that same one-sided topology, with the background relation pointing
```
tmadlener, line 433: should this be fixed in the producer? I.e. should we try to make a proper MC tree if possible? Also the changes above do seem to fill both directions? What am I missing?

test/k4FWCoreTest/src/components/ExampleFunctionalProducerMultiple.cpp
```diff
@@ -79,6 +79,11 @@ struct ExampleFunctionalProducerMultiple final : k4FWCore::Producer<retType()> {
     auto part1 = particles.create(1, 2, 3, 4.f, 5.f, 6.f, v, v, v);
     auto part2 = particles.create(2, 3, 4, 5.f, 6.f, 7.f);
 
+    // part2 is a daughter of part1, but its own parent list is left empty. That
+    // one-sided relation is what an incomplete relation table in a background
+    // file looks like, and overlaying has to keep it instead of dropping it.
```
tmadlener, line 84: Ah, maybe this "artifact" should be fixed in whatever produces the background events?

tmadlener (commented):
This looks like it could also be a bit of a performance improvement given that we replace a map lookup with a pure indexing. Did you by any chance measure that? I don't think it's necessary to do so explicitly, but if by chance the number exists, I would be interested if the lookup is in any way something costly in this whole overlay thing.

test/k4FWCoreTest/scripts/CheckOutputFiles.py
madbaron: It's L85 that does the one-sided filling.
The only thing we could change is go to a 3 particles record with: 1 with only daughters, 1 with only parents and 1 with both links, so that all cases are tested.

For this PR however what we have is sufficient, and a general update probably belongs to somewhere else (happy to note it in an issue, if you like)

test/k4FWCoreTest/src/components/ExampleFunctionalProducerMultiple.cpp
madbaron: Maybe, but a priori we have no control over what comes in, so we should be ready to handle "as properly as possible" any input?

madbaron:
> This looks like it could also be a bit of a performance improvement given that we replace a map lookup with a pure indexing. Did you by any chance measure that? I don't think it's necessary to do so explicitly, but if by chance the number exists, I would be interested if the lookup is in any way something costly in this whole overlay thing.

I was curious as well, so yesterday I tried to measure it for my education while setting up this PR:

## Setup

| | |
| --- | --- |
| Baseline | `main` at `5464018` |
| Change | PR #429 (`simplify_overlay_particle_relations`) |
| Build | `CMAKE_BUILD_TYPE=Release`, each revision built into its own Docker volume from its own git worktree |
| Container | `ghcr.io/muoncollidersoft/mucoll-sim-ubuntu24:post_3_1_test`, linux/arm64 |
| Peak RSS | `resource.getrusage(RUSAGE_CHILDREN).ru_maxrss` around the `k4run` subprocess |
| Wall time | median of 3 runs after a warm-up run |

Synthetic background frame: `N` MCParticles in a parent/daughter chain (every particle has one parent and one daughter, all relations symmetric), plus 50k SimTrackerHits pointing at them. Small signal event, one background event
overlaid per signal event, 5 events. Identical options file for both builds.

## Results

| N particles | peak RSS `main` | peak RSS #429 | saved | wall `main` | wall #429 | speed-up |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 50,000 | 431 MiB | 420 MiB | 11 MiB | 0.72 s | 0.54 s | 1.33x |
| 200,000 | 601 MiB | 561 MiB | 40 MiB | 1.65 s | 0.91 s | 1.81x |
| 500,000 | 934 MiB | 835 MiB | 99 MiB | 4.03 s | 2.06 s | 1.96x |

- Background merging is about twice as fast at the larger sizes. The wall times include roughly 0.3-0.5 s of fixed `k4run` and Gaudi start-up, so the merge step itself improves by more than the ratio suggests. 
- The saving is transient, not steady-state. The maps lived for the duration of one background frame, so this shows up in peak RSS only while a frame is being merged. Under the parallel background reading of the follow-up PR, where several frames are in flight at once, the per-frame saving multiplies.

k4FWCore/components/OverlayTiming.cpp
madbaron: You are right that both accesses are guaranteed by construction, and right that the cost is immeasurable here. I would still like to keep `at()`, because the downside is a bit worse than a crash.

podio collections are backed by a `std::deque`, and `at()` forwards to `std::deque::at`, so the bounds check is real. The flip side is that `std::deque::operator[]` out of range is undefined behaviour rather than a reliable crash. For a relation index that might mean silently attaching a hit to the wrong particle, which is the failure mode #425 was about.

Trading an unmeasurable branch for turning a silent-corruption bug into a `FATAL` that Gaudi reports seems worth it, so I have left the `at()` calls as they are.

k4FWCore/components/OverlayTiming.cpp
madbaron: Agreed on both counts.

I have added an `assert`, but at the point where the values are narrowed rather than inside the helper: the helper receives values that have already been narrowed, so if the cast wrapped, the damage is done before it is called and it cannot detect the actual problem. 

Note that the default build type (at least in the mucoll stack) is `RelWithDebInfo`, which carries `-DNDEBUG`, so the assert is compiled out of every build we normally run. It is live in a `Debug` build. 

I did not add a real runtime check because the condition is unreachable in practice. Overflowing would need more than 2.1e9 particles in a single output collection, so a job would be out of memory long beforehand.

k4FWCore/components/OverlayTiming.cpp
tmadlener: Fair enough. Forgot about the `std::deque` in the podio internals once again.

tmadlener:
Thanks for that benchmark. That speedup is more than I would have expected, but it probably heavily relies on the actual MCParticle tree / setup. In any case it effectively can't be slower and it will use less memory by construction since we eliminate a complete per-event map.

test/k4FWCoreTest/src/components/ExampleFunctionalProducerMultiple.cpp
tmadlener: Yes, especially since the input files where this caused issues already exist. Nevertheless, it would be nice to figure out why these files are as they are and whether that is something that can be fixed.

tmadlener (commented):
This looks good to me. Maybe @davidlange6 wants to have another looks since he first spotted / suggested this.

pre-commit failures are unrelated to these changes, I think. They are at least also present on `main` and likely due to the switch in stacks with different versions of `clang-tidy`.

madbaron:
Shall we get this merged?
(I'm happy to take care of any possible follow ups in a later PR)

tmadlener:
Unless someone still comments here, I will merge on Monday.

test/k4FWCoreTest/src/components/ExampleFunctionalProducerMultiple.cpp
jmcarcell:
```suggestion
```
This comment is irrelevant outside of this PR

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -286,45 +284,37 @@ retType OverlayTiming::operator()(const edm4hep::EventHeaderCollection& headers,
...
-        int j = oparticles.size();
-        for (size_t i = 0; i < bgParticles.size(); ++i) {
+        // The background particles are copied in order, so the copy of background
+        // particle i ends up at offset + i and no old-to-new index map is needed.
+        const int offset = static_cast<int>(oparticles.size());
+        const int nBgParticles = static_cast<int>(bgParticles.size());
+        // podio collections are limited to the positive range of int anyway, so
+        // this only documents that the narrowing above and the offset + index
+        // arithmetic below cannot overflow.
+        assert(oparticles.size() + bgParticles.size() <= static_cast<size_t>(std::numeric_limits<int>::max()));
```
jmcarcell, line 295:
```suggestion
```
The check doesn't run in Release builds, plus it would be a file with at least multiple GB (or tens?) per event, I'm sure there are other problems

madbaron: This assert was requested by @tmadlener in https://github.com/key4hep/k4FWCore/pull/429#discussion_r3971855126 (see discussion there).

tmadlener: My thinking process was: It's unlikely to happen here, but if it does I want it screaming at me while debugging. Hence, `assert` because that doesn't do anything in Release, but when I am debugging I am (almost certainly) not in a release build.

jmcarcell: I was thinking it's almost impossible, I get about 409 GB per 1e9 MCParticles without relations, so if we are at 4e9+ we probably have, at least, a memory problem.

tmadlener: Fair enough. Let's remove it then and hope that memory prices stay high enough so that we can never afford to overlay 1e9 MCParticles ;)

k4FWCore/components/OverlayTiming.cpp
```diff
@@ -286,45 +284,33 @@ retType OverlayTiming::operator()(const edm4hep::EventHeaderCollection& headers,
...
-        // To fix the relations we will need to have a map from old to new particle index
-        std::map<int, int> oldToNewMap;
-        std::map<int, std::pair<std::vector<int>, std::vector<int>>> parentDaughterMap;
-
         const auto& bgParticles = backgroundEvent.get<edm4hep::MCParticleCollection>(m_MCParticleCollectionName);
-        int j = oparticles.size();
-        for (size_t i = 0; i < bgParticles.size(); ++i) {
-          auto npart = bgParticles[i].clone(false);
+        // The background particles are copied in order, so the copy of background
+        // particle i ends up at offset + i and no old-to-new index map is needed.
```
tmadlener:
```suggestion
        // particle i ends up at offset + i
```
no information, only history how we came here

tmadlener (approved):

test/k4FWCoreTest/scripts/CheckOutputFiles.py
```diff
@@ -392,6 +392,15 @@ def check_metadata(filename, expected_metadata):
     ],
 )
 
+
+def mc_relation_indices(particle):
```
jmcarcell: Why was this added? This is doing nothing in the example below since the lists are empty or have one element?

madbaron: Honestly I don't remember anymore :)
It might be an erroneously ported remain from the two follow-up PRs to this one when I split this off. In these at some point I had introduced a more complex mc particle collection to test the various caveats of missing links. This looks like something related.

We can kick this out from here I think

test/k4FWCoreTest/scripts/CheckOutputFiles.py
```diff
@@ -392,6 +392,15 @@ def check_metadata(filename, expected_metadata):
...
+    """Indices of the parents and daughters of an MCParticle, as sorted lists."""
+    return (
+        sorted(p.getObjectID().index for p in particle.getParents()),
+        sorted(d.getObjectID().index for d in particle.getDaughters()),
+    )
+
+
```
jmcarcell:
```suggestion
```

test/k4FWCoreTest/scripts/CheckOutputFiles.py
```diff
@@ -411,6 +420,22 @@ def check_metadata(filename, expected_metadata):
     if particle_overlay_flags[n_signal_mc:] != [True] * n_background_mc:
         raise RuntimeError("Background particles should be flagged as overlay")
 
+    # The producer makes part2 a daughter of part1 without filling in part2's own
+    # parent list. Both blocks, signal and overlaid background, have to come out
+    # with that same one-sided topology, with the background relation pointing
+    # inside the background block only.
+    for block_start in (0, n_signal_mc):
+        relations = {i: mc_relation_indices(overlaid_particles[block_start + i]) for i in range(2)}
```
jmcarcell:
```suggestion
        relations = {i: overlaid_particles[block_start + i].getObjectID().index for i in range(2)}
```

test/k4FWCoreTest/scripts/CheckOutputFiles.py
jmcarcell: Removed then

jmcarcell: Ah nvm, I misunderstood what they do, bringing it back...

PR #430, opened 2026-09-10
Fix clangd NRVO warnings in location helpers

jmcarcell:
BEGINRELEASENOTES
 - Avoid clangd NRVO (named return value optimization, -Wnrvo, "not eliding copy on return") warnings in location helpers

ENDRELEASENOTES

### DIFF
--- a/k4FWCore/include/k4FWCore/Transformer.h
+++ b/k4FWCore/include/k4FWCore/Transformer.h
@@ -125,7 +125,7 @@ namespace details {
      * @param i  The index of the input
      * @return   A range of the input locations
      */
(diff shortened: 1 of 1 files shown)


tmadlener:
Out of curiosity, is NRVO now happening with naming a concrete type in the signature? And do you know why it behaves differently with `auto`?

jmcarcell:
It seems it's a limitation of clang (https://github.com/llvm/llvm-project/issues/95280). The type is the same in both cases

tmadlener (approved):
OK, interesting. I think it potentially also increases readability for new-comers (and maybe doxygen), so I would say we go ahead with this.

PR #431, opened 2026-09-19
Add --evt-max alias to k4run

jmcarcell:
BEGINRELEASENOTES
 - Add `--evt-max` as a k4run alias for the Gaudi event limit, to make it clear it's the same as `ApplicationMgr.EvtMax`

ENDRELEASENOTES

If both aliases are passed, argparse uses the last value provided.

@dudarboh


tmadlener (commented):
Do we already have it default to *run all events*? I guess we do, but is simply not setting `ApplicationMgr().EvtMax` where it happens?

jmcarcell:
It is always set: https://github.com/key4hep/k4FWCore/blob/main/python/k4FWCore/ApplicationMgr.py#L105 
Otherwise I guess the default from Gaudi (-1) applies if not set

tmadlener:
Maybe we can put the `-1` default into the python arguments as well explicitly? That would make the chase for figuring out what the default is and where it comes from a bit easier.

dudarboh:
When generating events using [PythiaInterface](https://github.com/key4hep/k4Gen/blob/80f6c6606a4cdc3174c05ac5a3ddf8e95aee0c9a/k4Gen/src/components/PythiaInterface.h)  the number of events to generate set in the pythia card settings is completely ignored it seems.
```
Main:numberOfEvents = 100         ! number of events to generate
```
So `--num-events -1` makes the generator run infinitely.

If one does not specify `--num-events` at all to the `ApplicatioMgr`, then some people could get a wrong impression that, then the value from the card will be used.

At least adding some info message to `PythiaInterface` would be a nice to avoid confusion.

tmadlener:
That might be something that needs fixing in `k4Gen`? E.g. https://github.com/key4hep/k4SimDelphes/blob/51aa8edaa5e0012519e211a600959e79eab5d625/standalone/src/DelphesPythia8Reader.h#L91 and then issuing an early stop in case that is reached.

The current default is already `-1`, so simply stating it explcitly shouldn't change behavior.

jmcarcell:
> Maybe we can put the `-1` default into the python arguments as well explicitly? That would make the chase for figuring out what the default is and where it comes from a bit easier.

No because the argument overrides anything in the steering files as it is now, I've put in the description.

jmcarcell:
> That might be something that needs fixing in `k4Gen`? E.g. [key4hep/k4SimDelphes@`51aa8ed`/standalone/src/DelphesPythia8Reader.h#L91](https://github.com/key4hep/k4SimDelphes/blob/51aa8edaa5e0012519e211a600959e79eab5d625/standalone/src/DelphesPythia8Reader.h#L91) and then issuing an early stop in case that is reached.
> 
> The current default is already `-1`, so simply stating it explcitly shouldn't change behavior.

Open an issue in k4Gen @dudarboh

tmadlener:
> No because the argument overrides anything in the steering files as it is now, I've put in the description.

Ah yes, forgot about that.

tmadlener (approved):

